PR #47, opened 2026-05-10
feat: detect and preserve Query/Header/Body/Path param sources

### LINKED ISSUE #14: Support Query() and Header() FastAPI parameter types
## Summary

The API inspector currently handles `Body()` parameters but doesn't distinguish `Query()` or `Header()` FastAPI parameter types. These should be detected and preserved through conversions.

## Current behavior

`Query()` and `Header()` params are treated as generic parameters, losing their semantic meaning.

## Expected behavior

- API inspector detects `Query(...)`, `Header(...)` parameter types
- ParameterInfo or ToolInfo captures the parameter source (body/query/header/path)
- API generator preserves the correct parameter type in output
- CLI/MCP generators handle these appropriately (e.g., query params → CLI options)

## Relevant files

- `src/intpot/core/inspectors/api.py`
- `src/intpot/core/models.py`
- `src/intpot/core/generators/api.py`
- `src/intpot/templates/api_app.py.j2`

MhussainD4772:
Closes #14

## What changed
- Added `ParamSource` enum to `models.py` (body/query/header/path)
- Added `param_source` field to `ParameterInfo`
- Added `_get_param_source()` to `api.py` to detect FastAPI param types
- Updated `api_app.py.j2` template to use correct FastAPI type in generated code
- Added test `test_query_and_header_param_sources` to verify detection works

## Testing
145 passed, 1 pre-existing failure unrelated to this change


MhussainD4772:
Hi, just a heads up : the initial CI run failed due to linting errors caught by ruff, not related to the logic changes. Specifically:

Unused import fastapi that was left in accidentally
Unsorted imports in api.py and the test file

Fixed by running uv run ruff check src/ tests/ --fix and uv run ruff format src/ tests/ and pushed the formatting fixes in the second commit. All 145 tests still passing locally.

tugrulguner:
Thank you, I think there is only one simple ruff error remained

MhussainD4772:
> Pushed a formatting fix  `models.py` wasn't included in the ruff format commit. All checks passing locally now (`ruff check` and `ruff format --check` both clean, 145 tests passing). Hopefully we should be all good

MhussainD4772:
Could you please approve the workflow

tugrulguner:
> Could you please approve the workflow

Of course, sorry, was a bit sick these days, taking time to do things

tugrulguner:
### Issues and suggestions

**1. Strip the “update narration” comments.** All four files have leftover commit-message-as-comment lines:
- `# Updated to pass down the param source`
- `# Updated to detect param source`
- `# Adding a parameter source enum`
- `# updated`
- `# Updated new test to test header and query param`

These describe the diff, not the code. Six months from now nothing is "updated" — just deleted. Please drop them before merge.

**2. Test coverage is thin — claims four sources, tests two.** The one new test only asserts `Query` and `Header`. Missing:
- **`ParamSource.body` regression** — existing path, but worth one assertion so the `Body` fallback in the template doesn't silently regress.
- **Explicit `Path(...)` detection** — covered by the mapping in `_get_param_source` but not asserted.
- **Implicit path-param fallback** — `def foo(item_id: int):` against `/items/{item_id}` should produce `ParamSource.path`. This is the new logic at the inspector's path-fallback branch and it has no test.
- **End-to-end generator output.** The template change is the most user-visible part of this PR and there's no test that the generated code (a) imports `Query, Header` and (b) emits `q: str = Query(...)` correctly. The Jinja namespace/set-collection idiom is fiddly enough that one snapshot-style test would catch a lot of future breakage.

**3. Template’s `.value | title` is fragile.** `param.param_source.value | title` works today because `body`/`query`/`header`/`path` all title-case cleanly to FastAPI class names. If anyone ever adds `cookie`, `form`, or a two-word value like `form_data`, it'll silently render as `Form_Data` and produce broken imports. Either:
- Add a `@property fastapi_class` on `ParamSource` that returns the exact class name, or
- Precompute a `fastapi_imports` list in the generator and pass it to the template so the template doesn't have to do string transforms at all.

Not a blocker — flag for follow-up.

**4. Template logic could move out of Jinja.** The namespace-based set building for FastAPI imports is doing real work in a template:

```jinja
{%- set ns = namespace(imports=[]) %}
{%- for tool in tools %}
  {%- for param in tool.parameters %}
    {%- if param.param_source %}
      {%- set src = param.param_source.value | title %}
      ...
```

This is the standard Jinja idiom but it's the kind of stateful loop that's much cleaner as a precomputed list on the generator side, passed as `fastapi_imports`. Fine as-is for now; worth a refactor when a fifth source type lands.

**5. Edge case (probably fine): `Query` on a path param.** If someone writes `def foo(item_id: int = Query(10)):` for `/items/{item_id}`, `_get_param_source` returns `ParamSource.query` and the path-param fallback never fires. FastAPI itself raises on this combination, so it's not a real-world concern, but worth noting that path-param status is currently not a hard override.

**6. Scope alignment with #14.** Issue #14 also asks for "CLI/MCP generators handle these appropriately (e.g., query params → CLI options)." This PR intentionally narrows to detect + preserve in the API path only. That's the right cut for a single PR, but let's file a follow-up issue for the CLI/MCP generator handling so it doesn't drop off the roadmap.

### Risk assessment

| | |
|---|---|
| **Correctness** | ✅ Core logic is right. Existing `Body` path is unchanged; new sources flow through cleanly. |
| **Backward compat** | ✅ `param_source = None` default + template fallback to `Body` mean every existing inspector and every existing API source keeps working. |
| **Performance** | ✅ One dict lookup per param. Negligible. |
| **Security** | ✅ Inspection-only, no user input handling. |
| **Test coverage** | ⚠️ Two of four documented sources tested; template output untested. Biggest gap in this PR. |
| **Style** | ⚠️ Update-narration comments need to come out before merge. |

### Verdict

Architectural shape is right and the implementation is clean. Before merge:

1. **Must:** Remove the five "Updated …" / "Adding …" comments.
2. **Should:** Add tests for `Body`, explicit `Path`, the implicit path-param fallback, and at least one generator-output assertion so the Jinja change is covered.
3. **Nice to have:** File a follow-up issue for CLI/MCP generator handling (the second half of #14) so it doesn't fall off the roadmap.

Once those land this is a clean, well-scoped contribution.

MhussainD4772:
Thank you for the deep insights, I'll work on these thru the weekend and raise a new PR!! Appreciate the advice.

Since I'm a junior dev this is great learning.

tugrulguner:
> Thank you for the deep insights, I'll work on these thru the weekend and raise a new PR!! Appreciate the advice.
> 
> Since I'm a junior dev this is great learning.

Hey, thank you for contributing :) you don't need to rise a new PR, continue with this one, push your fixes and changes to this existing branch

MhussainD4772:
Addressed the must-fix items and one of the should-fix items in the latest commit:
Must fix:

Removed all narration-style comments (# updated, # Updated new test...)
Added test coverage:

test_body_param_source : explicit Body() detection
test_path_param_source_explicit :  explicit Path() detection
test_path_param_source_implicit :  implicit path-param fallback when no annotation is used
test_generator_output_correct_fastapi_types :  end-to-end test asserting the generated code contains Query( and Header(


Should fix:

Replaced fragile .value | title template logic with a fastapi_class property on ParamSource. The enum now owns the class-name mapping explicitly, so adding values like cookie or form_data in the future won't silently break.

Follow-ups:

Issue 4 (moving template logic into the generator) -  happy to do it in a separate PR if you'd like
Issue 6 : let me know if you'd like me to open a follow up issue for CLI/MCP generator handling, or if you'd prefer to track it yourself

MhussainD4772:
Could you please accept the merge

tugrulguner:
### Follow-up review

Thanks for the quick turnaround on the previous round — most of the feedback is in.

### Addressed

- ✅ `ParamSource.fastapi_class` property added on the enum, and the template now uses it instead of `.value | title`. Exactly the right shape — no more silent breakage if a future source name doesn't title-case cleanly.
- ✅ Narration comments removed from `models.py` and `test_api_inspector.py`.
- ✅ Test coverage closed across the board:
  - `test_body_param_source` covers the `Body` regression path
  - `test_path_param_source_explicit` covers `Path(...)`
  - `test_path_param_source_implicit` covers the path-fallback branch (the new logic from this PR)
  - `test_generator_output_correct_fastapi_types` covers the Jinja template end-to-end

That last one is especially nice — the template logic was the highest-risk part of the change and now it's pinned down by an integration test that asserts both the import line and the per-param `Query(...)` / `Header(...)` rendering.

### Still outstanding

**1. Two narration comments survived in `src/intpot/core/inspectors/api.py`:**

```python
# Updated to pass down the param source
def _get_param_source(obj: Any) -> ParamSource | None:
```

```python
# Updated to detect param source
param_source = None
```

The third comment in that block — `# Fall back to the path if name is in path params and no FastAPI annotation` — is fine, it explains the *why* of the fallback. Just the two `# Updated to ...` lines should come out.

**2. `api_app.py.j2` lost its trailing newline.** Diff footer shows `\ No newline at end of file` on the last line. The repo's pre-commit config has `end-of-file-fixer` enabled, so this will fail `make check` / CI. Running `pre-commit run --all-files` locally (or just adding a newline back) will fix it.

### Verdict

Both items are one-line fixes. Once those land I'll merge — the substantive review is done.

MhussainD4772:
Both addressed in the latest commit:

Removed the two # Updated to ... comments in api.py (kept the # Fall back to path... one since it explains the why)
Restored the trailing newline in api_app.py.j2

Ran ruff check and ruff format --check locally — both clean. 149 tests passing.
Screenshot 2026-05-24 at 2 15 46 PM
Screenshot 2026-05-24 at 2 16 51 PM

tugrulguner:
Both items from the last round are in:

- ✅ The two `# Updated to ...` narration comments in `src/intpot/core/inspectors/api.py` are gone. The remaining `# Fall back to the path …` comment stays — that one explains *why* the fallback exists, which is genuinely useful.
- ✅ `api_app.py.j2` trailing newline restored — `end-of-file-fixer` will pass.

That clears everything from both review rounds. LGTM 🚢 — merging.

Thanks @MhussainD4772 for sticking with the back-and-forth and for the clean response on every iteration. The `fastapi_class` property and the end-to-end generator test in particular are nice touches above what the issue asked for.

I'll open a follow-up issue for the CLI/MCP generator side of #14 (query params → CLI options, etc.) so it doesn't get lost — would welcome another PR from you on that if you're up for it.

tugrulguner (approved):

MhussainD4772:
Thanks so much, really appreciate the thorough reviews learned a lot from the back-and-forth. Yes, I'd love to take on the CLI/MCP generator follow-up.

MhussainD4772:
Hello @tugrulguner could you guide me to the CLI/MCP generator issue that you opened so I can start working on that I cannot seem to find it.

tugrulguner:
> Hello @tugrulguner could you guide me to the CLI/MCP generator issue that you opened so I can start working on that I cannot seem to find it.

Hey, no worries, pick one from issues, also dont forget to start this repo if you like it :)

PR #83, opened 2026-08-15
fix: stop annotating None-returning tools as returning str

tugrulguner:
Consolidated review item **5**. Reproduced, and it is a 500 on every call to an affected endpoint.

## The failure

```python
@app.tool()
def log_it(msg: str, level: str = "info") -> None:
    """Returns nothing."""
    return None
```

`intpot eject --to api` generated `-> str`, and FastAPI validates the response against the annotation:

```
fastapi.exceptions.ResponseValidationError: 1 validation error:
  {'type': 'string_type', 'loc': ('response',), 'msg': 'Input should be a valid string', 'input': None}
```

A tool with **no** return annotation had the same problem the moment it returned anything non-string — `untyped(a, b) -> a + b` also 500'd.

## Cause

`python_type_name` answers both "no annotation" and `-> None` with `"str"`:

```python
if annotation is inspect.Parameter.empty or annotation is None:
    return "str"
```

That is a reasonable default for an unannotated **parameter** — a CLI argument really is a string. It is wrong for a return type, where the annotation is a contract FastAPI enforces.

## Fix

A separate `python_return_type_name` keeps the two apart:

| source | before | after |
|---|---|---|
| `-> None` | `str` | `None` |
| no annotation | `str` | `Any` |
| `-> int` | `int` | `int` |

`Any` is the honest answer when the source never said, and it tells FastAPI not to constrain the response. It was already in the generator's typing-import scan, so `from typing import Any` appears in the generated module — verified, since an annotation nothing imports is a `NameError` at import time.

Parameters keep using `python_type_name` and are unaffected.

## Why the existing tests passed

The live `serve --api` path never had this failure: `runtime_builders` does not derive a response model from `return_type`, so serving a `-> None` tool returns `200 null` even on `main`. Only **ejected and generated** code carries the annotation into FastAPI. The runtime tests were exercising the one path where the bug is invisible.

## Tests

`tests/test_return_types.py`, 10 cases, executing the generated app rather than reading it:

- the mapping itself, for all three input shapes
- `ToolInfo.return_type` distinguishes `None` from `Any`
- **an ejected `-> None` endpoint returns 200 and `null`** — the regression
- an ejected unannotated endpoint returns its actual value (`5`, not a 500)
- `from typing import Any` is present in the generated module
- all three targets still generate importable code

Four fail with the call sites reverted (keeping the new helper so the import resolves).

`make check`: **350 passed**.

### DIFF
--- a/changelog.d/83.fixed.md
+++ b/changelog.d/83.fixed.md
@@ -1,0 +1,4 @@
+Tools that return nothing are no longer annotated as returning `str`. An ejected FastAPI
+endpoint over a body returning `None` failed response validation on every call — a 500 —
+and a tool with no return annotation did the same as soon as it returned a non-string.
+`-> None` now stays `None`, and an unannotated return becomes `Any`.
--- a/src/intpot/core/inspectors/_utils.py
+++ b/src/intpot/core/inspectors/_utils.py
@@ -21,6 +21,25 @@ def python_type_name(annotation: Any) -> str:
     return text
 
 
+def python_return_type_name(annotation: Any) -> str:
+    """Name a tool's return annotation, keeping "nothing" distinct from "unknown".
+
+    `python_type_name` answers both with "str". That is a fair default for an
+    unannotated *parameter* and wrong for a return type: FastAPI validates the
+    response against this annotation, so `-> str` over a function returning
+    `None` is a 500 on every call rather than a type-checker complaint.
+
+    - `-> None` becomes `"None"`, which FastAPI accepts.
+    - No annotation at all becomes `"Any"`, which tells FastAPI not to constrain
+      the response — the only honest answer when the source never said.
+    """
+    if annotation is inspect.Parameter.empty:
+        return "Any"
+    if annotation is None or annotation is type(None):
+        return "None"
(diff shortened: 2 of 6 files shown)


tugrulguner (commented):
## Self-review

Nothing found against this one. Two things worth recording for whoever reads it later.

**Why the existing suite missed the bug.** `serve --api` never had this failure — `runtime_builders` doesn't derive a response model from `return_type`, so serving a `-> None` tool returns `200 null` even on `main`. Only ejected and generated code carries the annotation into FastAPI. The runtime tests were exercising the one path where the bug is invisible, which is why 340 tests passed over it.

**Relationship to #87.** They look like they overlap and don't. This changes what the *inspectors* and `runtime` report for `-> None` and unannotated returns; #87 computes the API target annotation from the transformed body and ignores `tool.return_type` entirely. Different code paths, no textual conflict, and I checked they compose: a `-> None` tool converted to API still gets its annotation from #87's outcome analysis, while an *ejected* one gets it from here.

Only file shared with another PR is `runtime.py` (with #88), and the two edits are in different functions.

PR #84, opened 2026-08-15
fix: preserve control flow when converting typer.echo to a return

tugrulguner:
Consolidated review item **4**. Reproduced — it produces a wrong answer silently, which is the worst failure mode this codebase has.

## The bug

Every `typer.echo` became its own `return`, regardless of position:

```python
# CLI source                          # converted to MCP (before)
for item in items:                    for item in items:
    typer.echo(item)                      return item
typer.echo('done')                    return 'done'
```

A command that printed every item returned **one**. No error, no warning. The same applied to echoes in both branches of a conditional, and any work after an echo became unreachable.

## Fix

Rewriting to a single `return` is only faithful when there is **exactly one echo and it is the last top-level statement**. That case is untouched — it's the common one, and it preserves the value's type:

```python
typer.echo(a + b)   →   return a + b        # still returns 5, not "5"
```

Everything else accumulates:

```python
_intpot_output = []
for item in items:
    _intpot_output.append(item)
_intpot_output.append('done')
return '\n'.join(str(_item) for _item in _intpot_output)
```

A bare `return` becomes the same join, so an early exit still returns what had been printed up to that point. Nested `def`s keep their own return semantics and are skipped, which the existing test for that already covered.

## Behaviour, verified by executing the generated module

| body | before | after |
|---|---|---|
| echo in a loop over `["a","b","c"]` | `"a"` | `"a\nb\nc"` |
| echo, then work, then echo | first echo only | `"start\n2"` |
| echo in each branch | first branch's value regardless | correct branch |
| echo, bare `return`, then a loop | returned at the first echo | `"empty"` early, `"a\nb"` otherwise |
| single trailing echo | `return a + b` | `return a + b` (unchanged) |

## Tests

Seven added to `tests/test_transforms.py`. Six of them **execute** the generated module — building the MCP server, registering the tools and calling the function — rather than reading the transformed source, because the old output was perfectly plausible as text.

The last one converts a multi-echo body to FastAPI, runs it, and asserts `{"result": "a\nb"}` over a real request.

Five fail on `main`.

`make check`: **347 passed**.

### DIFF
--- a/changelog.d/84.fixed.md
+++ b/changelog.d/84.fixed.md
@@ -1,0 +1,5 @@
+Converting a Typer command that prints more than once no longer loses output. Each
+`typer.echo` became its own `return`, so an echo inside a loop returned on the first
+iteration and a command that printed every item produced exactly one — silently. Bodies
+with several echoes, or echoes inside control flow, now collect their output and return
+all of it; a single trailing echo still returns its value directly.
--- a/src/intpot/core/transforms.py
+++ b/src/intpot/core/transforms.py
@@ -138,6 +138,7 @@ class _EchoToAccumulator(ast.NodeTransformer):
 
     def __init__(self) -> None:
         self._checker = _TyperEchoToReturn()
+        self._depth = 0
 
     def visit_Expr(self, node: ast.Expr) -> ast.AST:
         if not self._checker._is_typer_echo(node.value):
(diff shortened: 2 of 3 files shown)


tugrulguner (commented):
## Self-review

Reviewing my own PR before merge turned up a regression it introduced. Fixed in `4e1a4b4`; recording it here because the failure mode is worth keeping on the record.

### The regression

To protect a nested function's `return` statements, the transformer skipped nested `def`s entirely. That also skipped their `typer.echo` calls, so an echo inside a helper survived into the generated module — which never imports typer:

```python
def show(x):
    typer.echo(x)     # left untouched
```
```
NameError: name 'typer' is not defined
```

`main` did not have this: the old transformer visited every scope, so the nested echo became a `return` inside the helper. Wrong for the reasons this PR exists, but it ran.

### The fix

Descend into nested scopes and keep rewriting echoes, while tracking depth so only the converted function's *own* bare `return` becomes the joined output. Appending from a nested scope needs no `nonlocal` — the accumulator is mutated, never rebound:

```python
_intpot_output = []
def show(x):
    _intpot_output.append(x)
for item in items:
    show(item)
_intpot_output.append('done')
return '\n'.join(str(_item) for _item in _intpot_output)
```

`listing(["a", "b"])` → `"a\nb\ndone"`.

Covered by `test_an_echo_inside_a_helper_does_not_survive_into_the_output`, which asserts both that no `typer.echo` remains and that the generated module runs.

### Merge note

Conflicts textually with #87 — both edit `transforms.py` and `tests/test_transforms.py`. Different functions, so a mechanical rebase for whichever goes second.

tugrulguner (commented):
## Merge-order correction

I said in my earlier comment that this would conflict textually with #87 — both edit `transforms.py` and `tests/test_transforms.py`. I trial-merged all six open PRs onto `main` and **there are no conflicts at all**, in any order. My warning was wrong; the edits sit in different regions of both files.

The combined tree is also healthy, which clean merges alone don't prove:

- `make check` on all six merged: **406 passed**
- #84 and #87 compose correctly, which was the interaction I was most unsure about. A multi-echo body becomes an accumulator that always ends in a value return, so #87's analysis annotates it `dict` rather than `dict | None`, and the generated app answers `200 {"result":"a\nb"}` over a real request.
- #88's positional-only handling works in all three modes on the merged tree, and #83's `-> None` reporting is intact alongside it.

So these can go in whatever order suits you.

PR #85, opened 2026-08-15
fix: update installed agent guidance safely

tugrulguner:
## Summary

- generate parseable and relevance-activated instructions for Claude, Cursor, and Windsurf
- update Copilot and Codex guidance through bounded managed blocks without deleting ambiguous user content
- correct dry-run, installation, scaffolding, conversion-limit, and behavioral-verification guidance
- validate all six supported agent formats and warn when Codex instructions exceed its default 32 KiB limit

## Testing

- `make check` (Ruff, format, Pyright, 361 tests)
- parser-based YAML frontmatter tests
- managed-block update, partial repair, malformed-block, user-content preservation, and idempotency tests


tugrulguner (commented):
## Review

No blocking findings. 361 tests pass on the branch.

### Checked specifically

**The `pyyaml` addition.** A new dependency in a skills PR is worth a look — intpot's runtime deps are deliberately just `typer` and `jinja2`, and a user installing it shouldn't pay for a test's convenience. It's in `[dev]`, used by `tests/test_commands/test_add_skills.py` to `yaml.safe_load` the frontmatter. Correct placement: parsing the metadata for real beats asserting on substrings, and it costs users nothing.

**Appending to shared files.** The failure mode here is destroying a user's own `AGENTS.md` or `copilot-instructions.md`, which #72 was about. The tests cover updating in place without overwriting project content, which is the property that matters.

### Merge note

Overlaps three other PRs — `tests/test_skills_content.py` with #86 and #88, `templates/skills/intpot-python.md` with #88, `README.md` with #88. All mechanical, but whichever lands second needs a rebase.

tugrulguner (commented):
## Re-review — overlap with #88 resolved

`c06ec0b` removes the variadic-rejection paragraph and its test from this PR. I checked the content isn't lost in the handoff: `#88` carries both (`intpot-python.md:38` and `test_python_skill_warns_that_variadic_tools_are_rejected`), and this branch no longer mentions it. Correct owner — the rejection is #88's behaviour change.

Nothing further from me.

PR #86, opened 2026-08-15
fix: report conversion detection failures

tugrulguner:
## Summary

- translate single-file conversion detection failures into useful stderr messages and exit status 1
- create missing parent directories for single-file conversion output
- strengthen shipped-skill schema checks against `ToolInfo` and `ParameterInfo`
- correct internal guidance about `App.eject()`, Jinja filter failures, and runtime/ejected parity coverage

## Testing

- `make check` (Ruff, format, Pyright, 346 tests)
- regression coverage for all three conversion targets and nested output paths


tugrulguner (commented):
## Review

No findings. Small, focused, and it corrects something I got wrong.

### Verified

- 346 tests pass on the branch.
- The `AGENTS.md` correction is right and I checked it independently rather than taking it on trust:

  ```python
  jinja2.Environment().from_string('{{ x | nosuchfilter }}')
  # TemplateAssertionError: No filter named 'nosuchfilter'.
  ```

  My line said an unknown filter "renders blank". It doesn't — it raises at template load. I inherited that claim from the old `.claude/skills/pr-review` file in #69 and carried it across without testing it, which is exactly the failure the "verify against the source, not the previous wording" rule in `AGENTS.md` exists to prevent. Good catch.

- The `eject()` wording fix ("returns" rather than "writes") is also correct — it returns a string; the CLI command is what writes.

### What looks good

- Catching `DetectionError` and exiting through `typer.Exit(1)` with the message on stderr matches what every other command does, and the test asserts `stdout == ""` rather than just checking the exit code — that's the assertion that would catch a regression where the error leaks to stdout.
- `output.parent.mkdir(parents=True, exist_ok=True)` is the right place for it; the alternative of failing with a raw `FileNotFoundError` was the actual user-facing symptom.
- Softening the serve/eject parity claim is fair. That test does check representative body-parameter parity, not every signature shape — #88 is a live example of a shape it doesn't reach.

PR #87, opened 2026-08-15
fix: allow conditional FastAPI responses

tugrulguner:
## Summary

- derive generated FastAPI response annotations from reachable control-flow outcomes
- use `dict | None` when a converted function can return a wrapped value or fall through
- account for conditionals, exceptions, loops and loop `else` suites, context-manager suppression, match statements, and nested scopes
- execute a generated FastAPI app against both explicit-return and fallthrough request paths

## Testing

- `make check` (Ruff, format, Pyright, 352 tests)
- generated-app requests for both conditional outcomes
- focused control-flow outcome matrix


tugrulguner (commented):
## Review

No findings. I tried to break the outcome analysis and couldn't.

### Verified

Eleven control-flow shapes, checking the annotation each produces:

| body | annotation |
|---|---|
| `return a` | `dict` |
| `if a: return a` | `dict \| None` |
| `if/else`, both return | `dict` |
| `try/finally` with return | `dict` |
| `with ...: return a` | `dict` |
| `while True: return a` | `dict` |
| `raise` only | `None` |
| no return | `None` |
| loop, then return | `dict` |
| `try/except`, both return | `dict` |
| `try/except`, except falls through | `dict \| None` |

Every one is right, including the two I expected to catch it out — `while True` (no fallthrough because the loop can't exit normally) and `try/except` where only one arm returns.

Then end to end, since an annotation that type-checks can still fail at runtime:

```
a=5,b=1 → 200 {"result":5}
a=1,b=5 → 200 null
```

The second is the case that was a 500 before this PR.

### Notes

- Composing outcomes through `_then` rather than scanning for "is there a return anywhere" is the right shape — it's why `while True` and the loop-`else` cases come out correct instead of accidentally right.
- Treating an unguarded capture pattern as exhaustive in `match` is a nice touch; `case _:` alone would have been the easy version and would have been wrong for `case x:`.
- Complementary to #83, not overlapping: that one changes what the *inspectors* report for `-> None` and unannotated returns; this one computes the API target annotation from the transformed body and ignores `tool.return_type` entirely. Different code paths, no semantic clash — I checked both compose.

### Merge note

Conflicts textually with #84 — both edit `transforms.py` and `tests/test_transforms.py`. Different functions (`_target_return_type` vs `_from_cli`), so it's a mechanical rebase for whichever goes second.

tugrulguner (commented):
## Re-review — loop-`else` fix verified

Re-ran the outcome matrix with four loop-`else` cases added, since that's what `c3b0de1` targets:

| body | annotation |
|---|---|
| `for/else` where else returns | `dict` |
| `for/else` with a `break` in the body | `dict \| None` |
| `while/else` where else returns | `dict` |
| `for` body returns, else falls through | `dict \| None` |

All correct, along with the original eight. The `break` case is the subtle one — `break` skips the `else`, so the loop can finish without returning — and it comes out right.

352 tests pass. Nothing further from me.

PR #88, opened 2026-08-15
fix: preserve runtime signature semantics

tugrulguner:
## Summary

- preserve positional-only arguments across live Typer, FastAPI, and FastMCP execution
- preserve ordinary tool parameters named `self` and `cls`
- call synchronous and asynchronous positional-only parameters positionally from live FastAPI endpoints
- reject `*args` and `**kwargs` during registration instead of exposing inconsistent CLI, API, and MCP interfaces
- document the explicit variadic-signature contract in the shipped Python skill

## Testing

- `make check` (Ruff, format, Pyright, 349 tests)
- executed real Typer commands and FastMCP tool calls with positional-only arguments
- executed live FastAPI requests for sync and async positional-only tools
- executed live FastAPI requests for bound instance and async class methods
- executed an ejected FastAPI app with ordinary `self` and `cls` parameters


tugrulguner (commented):
## Review

Verified by running each mode, not by reading. The code and tests are sound; two things about scope.

### High: the fix covers one of the three modes, but the claim covers all of them

`changelog.d/88.fixed.md` says universal apps "now preserve positional-only ... parameters". Only the FastAPI path does.

```python
@app.tool()
def greet(name: str, greeting: str = "Hello", /) -> str:
    return f"{greeting}, {name}!"
```

| mode | result on this branch |
|---|---|
| FastAPI | `200 "Hello, a!"` ✅ |
| CLI (`build_typer_app`) | `exit 1` — `TypeError: greet() got some positional-only arguments passed as keyword arguments: 'name, greeting'` |
| MCP (`build_fastmcp_app`) | `TypeError`, same |

I checked `main`: both fail there identically, so this PR does not introduce them — it fixes FastAPI and leaves the other two. That's fine as a scope decision, but the changelog and PR title read as general.

Either narrow the wording to FastAPI, or apply the same `call_arguments` restoration in `build_typer_app` and `build_fastmcp_app`. The helper is already written; it's the call sites that don't use it.

`tests/test_runtime_builders.py` asserts serve/eject parity, and this is exactly the kind of signature shape that parity check doesn't reach — worth a case there either way.

### Medium: rejecting variadics is a breaking change filed as a fix

`@app.tool()` on a function with `*args` or `**kwargs` now raises `ValueError` at registration, which is import time for most apps. Turning silent misbehaviour into a loud failure is the right call, but it will break working-looking code on upgrade.

`changelog.d/88.fixed.md` should be `88.changed.md` so it lands under **Changed** in the release notes rather than being read as a bug fix.

### What looks good

- Restoring positional-only args via a closure over the parameter names, rather than reordering at the schema layer, keeps the wrapper honest about the original callable.
- Preserving ordinary parameters named `self`/`cls` is the right trade: a decorated method in a class body was never callable through this path anyway.
- The variadic error message names the tool, the parameter, and why — that's the difference between a rejection someone can act on and one they file a bug about.

tugrulguner:
Addressed both scope findings:

- Positional-only support is now genuinely universal across the live builders. Typer and FastMCP receive framework-facing signatures with named parameters, while the wrappers restore those values positionally before invoking the source function. FastAPI keeps its existing positional restoration.
- Added real execution regressions: a `CliRunner` command and `FastMCP.call_tool()` both exercise a positional-only parameter with a default. The existing real FastAPI requests remain.
- Renamed `changelog.d/88.fixed.md` to `changelog.d/88.changed.md` because rejecting variadic signatures is intentionally a behavioral change.

Verification after the changes: `make check` passes Ruff, formatting, Pyright, and all 349 tests.

tugrulguner (commented):
## Re-review — both findings addressed

Re-ran the exact repro from my last comment:

| mode | before | now |
|---|---|---|
| FastAPI | `200 "Hello, a!"` | `200 "Hello, a!"` |
| CLI | `exit 1`, `TypeError` | `exit 0`, `Hello, a!` |
| MCP | `TypeError` | `'Hello, a!'` |

And the changelog fragment is now `88.changed.md`, so the variadic rejection lands under **Changed** rather than reading as a bug fix.

The added coverage goes further than I asked for — positional-only across all three builders plus bound instance and class methods, which is the shape most likely to regress silently since `inspect.signature` treats bound and unbound differently.

349 tests pass on the branch. Nothing further from me.

PR #100, opened 2026-08-21
fix: carry dotted imports into generated code

tugrulguner:
Second in the agreed order, and deliberately before the dependency-closure work since it lands in the same extraction utility.

## The bug

`import os.path` binds **`os`** in the module namespace, not `os.path`. The scan recorded the dotted path as the binding:

```python
names = {alias.asname or alias.name for alias in node.names}   # {"os.path"}
```

A body using it says `os.path.join(...)`, whose root name is `os`. `{"os.path"} & {"os"}` is empty, so the import was dropped:

```python
import typer
import json                 # kept — binds `json`
                            # `import os.path` missing
```

```
$ python out.py where hello
NameError: name 'os' is not defined
```

The generated file imports and compiles cleanly; it fails when the command runs.

## Why it survived this long

The two adjacent forms were always correct:

| form | binds | recorded | |
|---|---|---|---|
| `import os.path` | `os` | `os.path` | ✗ dropped |
| `import os.path as p` | `p` | `p` | ✓ |
| `from os import path` | `path` | `path` | ✓ |

Only the plain dotted form was wrong, and aliased/`from` imports are what most code uses.

## Fix

Bindings are computed per statement type: `ImportFrom` binds the alias name; `Import` binds the alias name, or its first dotted segment when unaliased.

## Tests

Six in `tests/test_inspectors/test_utils.py`:

- dotted import kept when used — the direct regression
- aliased and `from` forms unchanged, so the fix can't quietly alter them
- **an unused dotted import is still dropped** — the filter must not degrade into "keep everything", which is the easy wrong fix here
- `import a.b.c` does not match a body mentioning `b` or `c`, only `a`
- a generated command using a dotted import **runs and produces the right output** — the failure was at runtime, so the test reaches runtime

Two fail on `main`: the direct case and the runtime one. The other four are guards against over-correcting.

`make check`: **416 passed**.

### DIFF
--- a/src/intpot/core/inspectors/_utils.py
+++ b/src/intpot/core/inspectors/_utils.py
@@ -40,6 +40,23 @@ def python_return_type_name(annotation: Any) -> str:
     return python_type_name(annotation)
 
 
+def _bound_names(node: ast.Import | ast.ImportFrom) -> set[str]:
+    """The names an import statement actually binds in the module namespace.
+
+    `import os.path` binds `os`, not `os.path` — the body that uses it says
+    `os.path.join(...)`, whose root name is `os`. Recording the dotted path
+    meant the binding never matched anything the body referenced, so the import
+    was dropped and the generated module raised `NameError: name 'os' is not
+    defined` at runtime.
+
+    `import os.path as p` binds `p`, and `from os import path` binds `path`;
+    both are already the full alias name.
+    """
+    if isinstance(node, ast.ImportFrom):
+        return {alias.asname or alias.name for alias in node.names}
+    return {alias.asname or alias.name.split(".", 1)[0] for alias in node.names}
+
+
 def extract_source_imports(fn: Any) -> list[str]:
     """Extract imports from the source file that are referenced in the function body.
 
@@ -64,8 +81,7 @@ def extract_source_imports(fn: Any) -> list[str]:
     import_nodes: list[tuple[ast.stmt, set[str]]] = []
     for node in module_tree.body:
         if isinstance(node, (ast.Import, ast.ImportFrom)):
-            names = {alias.asname or alias.name for alias in node.names}
-            import_nodes.append((node, names))
+            import_nodes.append((node, _bound_names(node)))
 
     if not import_nodes:
         return []
--- a/src/intpot/core/inspectors/_utils.py
+++ b/src/intpot/core/inspectors/_utils.py
@@ -40,8 +40,8 @@ def python_return_type_name(annotation: Any) -> str:
     return python_type_name(annotation)
 
 
-def _bound_names(node: ast.Import | ast.ImportFrom) -> set[str]:
-    """The names an import statement actually binds in the module namespace.
+def _bound_name(node: ast.Import | ast.ImportFrom, alias: ast.alias) -> str:
+    """The name one alias binds in the module namespace.
 
     `import os.path` binds `os`, not `os.path` — the body that uses it says
     `os.path.join(...)`, whose root name is `os`. Recording the dotted path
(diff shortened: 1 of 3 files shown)


src/intpot/core/inspectors/_utils.py
```diff
@@ -64,8 +81,7 @@ def extract_source_imports(fn: Any) -> list[str]:
     import_nodes: list[tuple[ast.stmt, set[str]]] = []
     for node in module_tree.body:
         if isinstance(node, (ast.Import, ast.ImportFrom)):
-            names = {alias.asname or alias.name for alias in node.names}
-            import_nodes.append((node, names))
+            import_nodes.append((node, _bound_names(node)))
```
tugrulguner: Blocking: this still treats a multi-alias import as one indivisible statement. For `import os.path, typer`, `_bound_names()` returns `{"os", "typer"}` and a tool using `os.path` retains the full statement. `_collect_extra_imports()` then sees `typer` anywhere in that rendered string and drops the whole statement, including `os.path`; generated CLI/API/MCP code fails with `NameError`. Please split imports into individually tracked aliases before framework filtering (or parse/filter aliases individually in the renderer), and add an executed generated-target regression for this mixed statement. The changelog's broad claim is accurate once this case is covered.

tugrulguner (commented):
Blocking correctness issue: mixed import statements can still lose a required dotted import during rendering. See the inline comment. The standalone dotted-import behavior and the rest of the test coverage look good.

tugrulguner (commented):
Confirmed and fixed in `d4a2e04`. You were right, and the changelog claim was too broad.

Reproduced first:

```python
import os.path, typer
...
    return os.path.join("/tmp", name)
```
```
generated:  import typer          # `import os.path` gone
run:        NameError: name 'os' is not defined
```

Exactly your trace — `_bound_names()` returned `{"os", "typer"}`, the statement was retained, and `_collect_extra_imports()` matched `typer` in the rendered string and dropped it whole.

## Fix

Took the first option: split before filtering. `_split_aliases()` turns each `Import`/`ImportFrom` into one single-alias statement per bound name, so the pairing becomes `(statement, one name)` rather than `(statement, set-of-names)`. Each name is then kept or dropped on its own merit, and the renderer needs no alias parsing.

Filtering in the renderer would have worked too, but it would put import semantics in the template layer and leave the inspector still reporting a binding set it can't act on.

One deliberate side effect: `from x import a, b` where the body only uses `a` now emits `from x import a`, not both. Narrower, and it stops generated files carrying unused imports.

## Tests

Three added, all failing on the previous commit:

- `test_a_mixed_import_statement_keeps_the_name_the_body_uses` — `import os.path, typer` → `["import os.path"]`
- `test_a_mixed_from_import_keeps_only_the_name_in_use`
- `test_a_mixed_import_survives_into_a_running_command` — the executed regression you asked for: converts to CLI, `exec`s the module, invokes the command, asserts `/tmp/hello`

The changelog fragment now describes both failure modes rather than claiming dotted imports generally.

`make check`: 419 passed.

tugrulguner (commented):
Re-reviewed head `97e862c`. The mixed-import blocker is fixed at the right layer: aliases are split during inspection, so required non-framework imports survive downstream framework filtering while unused aliases are omitted. The added tests cover ordinary mixed imports, mixed `from` imports, and the executed regression.

Verification:
- focused inspector tests: 16 passed
- full `make check`: 419 passed; Ruff, formatting, and Pyright clean
- generated CLI, FastAPI, and FastMCP targets with `import os.path, typer` all executed successfully
- all GitHub checks pass across Python 3.11–3.13

No remaining blocking findings. Ready to merge.

PR #101, opened 2026-08-21
fix: report import failures instead of dumping a traceback

tugrulguner:
Findings from the codebase review, plus three rounds of review on this PR. Everything reproduced before being fixed.

The theme is one question: **what happens when a source can't be used?** Every answer was wrong in a different way.

## 1. Import failures dumped a traceback

`detect_source` executes the user's module. Only `DetectionError` was caught, so anything else propagated:

```
│ ❱ 1 import nonexistent_module_xyz                    │
╰─────────────────────────────────────────────────────╯
ModuleNotFoundError: No module named 'nonexistent_module_xyz'
```

`inspect`, `to *`, `eject` and `serve` all did this, exiting **2** (Typer's crash code). Directory scanning handled the identical file correctly, so the two paths disagreed — and the one people use most was the broken one.

**Fix.** Every command funnels through `_import_module_from_path`, so the wrap goes there. `SourceImportError` **subclasses** `DetectionError`, so each command's existing handler reports it cleanly with no per-command change.

That subclassing has a sharp edge, documented on the class and covered by tests in both directions: `discovery.py` must catch `SourceImportError` **before** `DetectionError`, or an import failure matches the parent, looks like "not an app", and goes silent — undoing #59.

## 2. `sys.exit()` at import was worse than a traceback

`SystemExit` derives from `BaseException`, so `except Exception` never saw it. The source's exit code became **intpot's**, with no output at all:

```
$ intpot inspect exiter.py     # module calls sys.exit(3)
$ echo $?
3
```

The directory scan did the same — exit 3, run over. `argparse.parse_args()` hits this on a bad parse.

Caught explicitly alongside `Exception` rather than widening to `BaseException`; a test asserts Ctrl-C still propagates.

## 3. Malformed Python was reported as "no app here"

`_has_framework_app_ast` swallowed `SyntaxError` and returned `False`, so a file that doesn't parse got the same message as a valid module containing no app — and a scan dropped it silently. Broken is not the same as uninteresting.

Parse failures now raise `SourceImportError` too, reusing the handling above. Valid non-app files stay quiet, in both single-file and directory form, with tests for each.

## 4. Install hints named the wrong package

`"fastapi" in module` matched a user's own missing `myfastapi_helper` and sent them to install `intpot[api]` — which wouldn't help and hid the real cause. Top-level name matched exactly now.

## 5. `load()`'s documented contract broke

Wrapping made `except ModuleNotFoundError` unreachable for path sources, contradicting `load()`'s own docstring. It now unwraps a `SourceImportError` caused by a `ModuleNotFoundError` and re-raises that, preserving `.name`.

`load()` **raises** — it's the Python API. The CLI commands exit 1; `load()` does not.

## 6. The unwrap created a circular cause chain

Introduced while fixing 5, caught in review:

```
exc.__cause__.__cause__ is exc  →  True
```

`raise original from e` set `original.__cause__ = e` while `e.__cause__` was already `original`. Anything walking the chain loops. Branch on identity now: an unchanged original is re-raised `from None`; a newly mapped error chains from what actually failed.

## 7. A tool parameter named `_fn` broke `serve --cli`

`build_typer_app` bound the tool as a default argument, putting it in the wrapper's own signature. Typer passed the user's `_fn="hello"` and the wrapper called a string. Bound by closure now.

## Tests

**35** in `tests/test_import_failures.py` plus one in `test_runtime_builders.py`.

Beyond the direct regressions, the ones that matter most are the ones asserting things *keep* working:

- valid non-app files stay quiet, single-file **and** in a scan
- a scan reports the bad file **and still converts the good one** — that assertion is what proves it continued, rather than merely not crashing
- `KeyboardInterrupt` still propagates
- the cause chain is walked with a seen-set, failing on a repeat rather than hanging

Verification note: these can't be run against `main` to show a red-to-green count, because they import a symbol that doesn't exist there. Individual fixes were each confirmed by reverting the relevant source file — 2 fail without the parse fix, 5 without the `SystemExit` and contract fixes, 1 without the cycle fix.

`make check`: **448 passed**.

### DIFF
--- a/src/intpot/converter.py
+++ b/src/intpot/converter.py
@@ -6,7 +6,7 @@ import functools
 from pathlib import Path
 from typing import Any
 
-from intpot.core.detector import detect_instance, detect_source
+from intpot.core.detector import SourceImportError, detect_instance, detect_source
 from intpot.core.models import SourceType, ToolInfo
 
 
@@ -102,6 +102,29 @@ class IntpotApp:
         return out.resolve()
 
 
+def _missing_module_error(exc: ModuleNotFoundError) -> ModuleNotFoundError:
+    """Point at the right extra, or hand the original error back unchanged.
+
+    The top-level module is matched exactly. `"fastapi" in module` also matched
+    a user's own missing `myfastapi_helper`, sending them to install
+    intpot[api] — which would not have helped and hid the real cause.
+    """
+    top_level = (exc.name or "").split(".", 1)[0]
+    if top_level == "fastmcp":
+        return ModuleNotFoundError(
+            "FastMCP is required for MCP support. "
+            "Install it with: pip install intpot[mcp]",
+            name=exc.name,
+        )
+    if top_level in ("fastapi", "uvicorn"):
+        return ModuleNotFoundError(
+            "FastAPI is required for API support. "
+            "Install it with: pip install intpot[api]",
+            name=exc.name,
+        )
+    return exc
+
+
 def load(source: Any) -> IntpotApp:
     """Load a source for conversion.
 
@@ -124,19 +147,14 @@ def load(source: Any) -> IntpotApp:
 
         source_type, app_instance = detect_instance(source)
         return IntpotApp(source_type, app_instance)
-    except ModuleNotFoundError as e:
-        # Match the top-level module exactly. `"fastapi" in module` also matched
-        # a user's own missing `myfastapi_helper`, sending them to install
-        # intpot[api] — which would not have helped and hid the real cause.
-        top_level = (e.name or "").split(".", 1)[0]
-        if top_level == "fastmcp":
-            raise ModuleNotFoundError(
-                "FastMCP is required for MCP support. "
-                "Install it with: pip install intpot[mcp]"
-            ) from e
-        if top_level in ("fastapi", "uvicorn"):
-            raise ModuleNotFoundError(
-                "FastAPI is required for API support. "
-                "Install it with: pip install intpot[api]"
-            ) from e
+    except SourceImportError as e:
+        # detect_source wraps whatever the user's module raised so the CLI can
+        # report it. This is the Python API, which documents ModuleNotFoundError
+        # — unwrap it back so callers catching that keep working.
+        if isinstance(e.__cause__, ModuleNotFoundError):
+            raise _missing_module_error(e.__cause__) from e
         raise
+    except ModuleNotFoundError as e:
+        # The live-instance path: inspectors import fastmcp/fastapi themselves,
+        # and that never goes through detect_source.
+        raise _missing_module_error(e) from e
--- a/src/intpot/converter.py
+++ b/src/intpot/converter.py
@@ -151,8 +151,16 @@ def load(source: Any) -> IntpotApp:
         # detect_source wraps whatever the user's module raised so the CLI can
         # report it. This is the Python API, which documents ModuleNotFoundError
         # — unwrap it back so callers catching that keep working.
-        if isinstance(e.__cause__, ModuleNotFoundError):
-            raise _missing_module_error(e.__cause__) from e
+        original = e.__cause__
+        if isinstance(original, ModuleNotFoundError):
+            mapped = _missing_module_error(original)
+            if mapped is original:
+                # Handing the original back means exactly that. `raise original
+                # from e` would set original.__cause__ = e while e.__cause__ is
+                # already original, so the chain points at itself and anything
+                # walking __cause__ loops forever.
+                raise mapped from None
+            raise mapped from original
         raise
     except ModuleNotFoundError as e:
         # The live-instance path: inspectors import fastmcp/fastapi themselves,
--- a/src/intpot/converter.py
+++ b/src/intpot/converter.py
@@ -125,13 +125,16 @@ def load(source: Any) -> IntpotApp:
         source_type, app_instance = detect_instance(source)
         return IntpotApp(source_type, app_instance)
     except ModuleNotFoundError as e:
-        module = e.name or ""
-        if "fastmcp" in module:
+        # Match the top-level module exactly. `"fastapi" in module` also matched
+        # a user's own missing `myfastapi_helper`, sending them to install
+        # intpot[api] — which would not have helped and hid the real cause.
+        top_level = (e.name or "").split(".", 1)[0]
+        if top_level == "fastmcp":
             raise ModuleNotFoundError(
                 "FastMCP is required for MCP support. "
                 "Install it with: pip install intpot[mcp]"
             ) from e
-        if "fastapi" in module:
+        if top_level in ("fastapi", "uvicorn"):
             raise ModuleNotFoundError(
                 "FastAPI is required for API support. "
                 "Install it with: pip install intpot[api]"
--- a/src/intpot/core/detector.py
+++ b/src/intpot/core/detector.py
@@ -31,8 +31,14 @@ class SourceImportError(DetectionError):
 _EXTRA_FOR_MODULE = {"fastmcp": "mcp", "fastapi": "api", "uvicorn": "api"}
 
 
-def _import_failure_message(source_path: Path, exc: Exception) -> str:
+def _import_failure_message(source_path: Path, exc: BaseException) -> str:
     """Explain an import failure in terms the user can act on."""
+    if isinstance(exc, SystemExit):
+        return (
+            f"Cannot import {source_path}: it called sys.exit({exc.code!r}) while "
+            f"being imported. Detection has to import the file, so module-level "
+            f"exits run too — guard them with `if __name__ == '__main__':`."
+        )
     detail = f"Cannot import {source_path}: {type(exc).__name__}: {exc}"
     missing = (
         getattr(exc, "name", None) if isinstance(exc, ModuleNotFoundError) else None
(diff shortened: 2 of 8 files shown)


src/intpot/core/detector.py
```diff
@@ -59,6 +93,10 @@ def _import_module_from_path(source_path: Path) -> Any:
     sys.modules[unique_name] = module
     try:
         spec.loader.exec_module(module)
+    except Exception as exc:
```
tugrulguner: Blocking: `SystemExit` derives from `BaseException`, not `Exception`, so a source that calls `sys.exit()` during import bypasses `SourceImportError`. Single-file commands inherit the source's arbitrary exit code without the promised diagnostic, and directory discovery can terminate the entire scan instead of reporting and continuing. Please catch `SystemExit` explicitly alongside `Exception` (do not broadly catch `BaseException`, so `KeyboardInterrupt` still propagates) and add regressions for both a single-file command and a directory scan.

src/intpot/converter.py
```diff
@@ -125,13 +125,16 @@ def load(source: Any) -> IntpotApp:
         source_type, app_instance = detect_instance(source)
         return IntpotApp(source_type, app_instance)
     except ModuleNotFoundError as e:
```
tugrulguner: Blocking public API regression: for path-based `intpot.load()`, `_import_module_from_path()` now wraps `ModuleNotFoundError` in `SourceImportError`, making this existing handler unreachable. That breaks the documented `ModuleNotFoundError` contract above and callers that catch it. Please either preserve the contract by handling/unwrapping a `SourceImportError` caused by `ModuleNotFoundError`, or explicitly document, export, and test the new public exception behavior. Preserving the existing contract is safer.

tugrulguner (commented):
Three blocking correctness/API-contract issues remain. Two are inline. The third is in `_has_framework_app_ast()` (`src/intpot/core/detector.py:61-65`): it still catches `SyntaxError` and returns `False`, so malformed source is reported as having no app rather than naming the parse error. That bypasses the new reporting path and conflicts with the changelog claim that source errors now print what went wrong. Please distinguish an unparseable source from a valid non-app file and add single-file plus directory-scan regressions.

The `_fn` closure fix, discovery catch ordering, and exact framework-name matching otherwise look sound. Local focused tests passed (32), full `make check` passed (435), and CI is green.

tugrulguner (commented):
Both confirmed and fixed in `8f3c1d2`. Reproduced each before changing anything.

## `SystemExit` — worse than described

You called it as bypassing `SourceImportError`. It's more than that: the source's exit code became **intpot's** exit code, with no output whatsoever.

```
$ intpot inspect exiter.py        # module calls sys.exit(3)
                                  # (nothing printed)
$ echo $?
3
```

And the directory scan did the same — exit 3, whole run over, so one bad file still kills the scan. That is exactly the failure #59 was meant to end, reintroduced through a path #59 never covered.

Fixed by catching `SystemExit` alongside `Exception`, listed explicitly rather than widening to `BaseException`. `test_keyboard_interrupt_is_not_swallowed` asserts Ctrl-C still propagates, since that's the risk of the lazy version of this fix.

The message names the code and the cause, because the trigger is nearly always an unguarded module-level exit — `argparse.parse_args()` on intpot's argv being the common one:

```
Cannot import /path/exiter.py: it called sys.exit(3) while being imported.
Detection has to import the file, so module-level exits run too — guard them
with `if __name__ == '__main__':`.
```

Both regressions added: single-file exits 1 with the message, and the directory scan reports `SKIP (import failed)` **and still converts the good file in the same directory** — that last assertion is the one that proves the scan continued rather than merely not crashing.

## `load()` contract — you were right, and I took the safer option

Wrapping made `except ModuleNotFoundError` unreachable for path sources, so `load()` stopped honouring the contract in its own docstring.

Preserved rather than redocumented, as you suggested. `load()` unwraps a `SourceImportError` whose `__cause__` is a `ModuleNotFoundError` and re-raises that — including `.name`, so callers inspecting the attribute keep working, not just those matching the type.

The install-hint logic moved into `_missing_module_error()`, shared by both paths. The live-instance path never goes through `detect_source` and still needs it, so leaving it inline would have meant two copies drifting.

```python
load(path_with_missing_module)   →  ModuleNotFoundError: No module named 'totally_absent_module'
                                    .name == "totally_absent_module"
```

Four tests: the contract for a path source, the hint still applied, an unrelated module handed back **unchanged** (identity-asserted, so it can't be quietly rewrapped), and `DetectionError` still raised for a non-app.

## Verification

Five of the new tests fail with the two source files reverted. `make check`: 442 passed.

tugrulguner (commented):
Two blocking issues remain after re-review. One is inline. The other is the previously reported syntax-error path at `src/intpot/core/detector.py:67-71`: `_has_framework_app_ast()` still catches `SyntaxError` and returns `False`, so malformed source is reported as having no app rather than naming the parse error. Please preserve/report the parse exception, ensure directory discovery reports it instead of treating it as an ordinary non-app, and add single-file plus directory-scan regressions while keeping valid non-app files quiet.

The `SystemExit` fix and its `KeyboardInterrupt`/continued-scan coverage are good. Focused tests passed (39), full `make check` passed (442), and CI is green.

src/intpot/converter.py
```diff
@@ -124,16 +147,14 @@ def load(source: Any) -> IntpotApp:
...
-            raise ModuleNotFoundError(
-                "FastAPI is required for API support. "
-                "Install it with: pip install intpot[api]"
-            ) from e
+    except SourceImportError as e:
+        # detect_source wraps whatever the user's module raised so the CLI can
+        # report it. This is the Python API, which documents ModuleNotFoundError
+        # — unwrap it back so callers catching that keep working.
+        if isinstance(e.__cause__, ModuleNotFoundError):
+            raise _missing_module_error(e.__cause__) from e
```
tugrulguner: Blocking: when `_missing_module_error()` returns the original unrelated `ModuleNotFoundError`, this creates a circular cause chain. `SourceImportError.__cause__` already points to that exception; raising the same exception `from e` makes its `__cause__` point back to the wrapper. I verified `exc.__cause__.__cause__ is exc` is `True`. This does not hand the public exception back unchanged and can break cause-chain traversal. Please branch on identity: re-raise an unchanged original from `None`, and raise a newly mapped framework error from the original missing-module exception. Add a regression through `load(path)` asserting there is no cycle; the direct helper identity test cannot detect the cycle introduced here.

tugrulguner (commented):
Confirmed and fixed in `c7e9b41`. Verified the cycle exactly as you described before changing anything:

```
exc                      ModuleNotFoundError
exc.__cause__            SourceImportError
exc.__cause__.__cause__  ModuleNotFoundError
exc.__cause__.__cause__ is exc  →  True
```

A defensive walk with a seen-set hit a repeat after 2 steps. So my "hands the original back unchanged" claim was false — the `raise ... from e` mutated the object I said I was passing through.

Fixed by branching on identity, as you specified:

```python
mapped = _missing_module_error(original)
if mapped is original:
    raise mapped from None      # unchanged really means unchanged
raise mapped from original      # newly mapped: what actually failed stays one link away
```

`from None` rather than a bare `raise` because the original needs its `__cause__` left clear; a bare re-raise inside the `except` would set `__context__` to the `SourceImportError` and produce the same confusing chain in a traceback, just via a different attribute.

After:

```
unrelated missing module: chain=['ModuleNotFoundError'] cyclic=False name='totally_absent_module'
```

## On the test

Your point that the identity test couldn't detect this was the useful part — I'd have written the regression at the helper level again and it would have passed. The cycle is created by the `raise`, not by the mapping, so the test has to go through `load(path)`.

`test_load_does_not_build_a_circular_cause_chain` does that and walks `__cause__` with a seen-set, failing on a repeat rather than hanging — a naive `while node.__cause__` assertion would have looped forever instead of reporting. It fails on the previous commit.

`make check`: 444 passed.

---

Worth noting this is three rounds on one PR, each finding real. The first two were edges I reasoned past instead of executing; this one I introduced while fixing the second. I've stopped treating "the tests pass" as sufficient for the exception-handling paths in particular — they're the code least likely to be exercised by the happy path I keep verifying.

tugrulguner (commented):
The circular exception-chain fix is correct, but the malformed-source blocker remains unchanged on `8239f50`.

I executed `intpot inspect` against malformed Python on this head. It exited 1 but printed only:

```
No FastMCP, Typer, or FastAPI app instance found in .../syntax.py
```

`_has_framework_app_ast()` still catches `SyntaxError` and returns `False` (`src/intpot/core/detector.py:67-71`), so the actual parse failure is hidden and directory discovery treats the file as an ordinary non-app. Please preserve/report the parse exception, add single-file and continued-directory-scan regressions, and keep valid non-app files quiet.

Non-blocking PR-description cleanup: it still says 16 tests / 435 passed and says `converter.load` “exits 1.” The current suite has 444 tests, and `load()` is a Python API that raises rather than exits.

Everything else now checks out: focused tests passed (41), full `make check` passed (444), the missing-module cause chain is no longer cyclic, and CI is green.

tugrulguner (commented):
Confirmed and fixed in `a91f4d7`. You were right that the earlier round left this untouched — I fixed the cycle and never went back to the malformed-source case.

Reproduced on the previous head:

```
$ intpot inspect syntax.py
No FastMCP, Typer, or FastAPI app instance found in .../syntax.py     ← identical to a valid non-app

$ intpot to cli dir -o out
Generated CLI app: out/good_cli.py                                     ← the malformed file: no mention
```

Same "quiet when it should not be" failure as #59, through a path #59 never touched.

## Fix

A parse failure raises `SourceImportError`, reusing the handling added earlier in this PR rather than adding a third error type. Importing a malformed module raises `SyntaxError` in Python too, so the scan's `SKIP (import failed)` label stays accurate.

```
single file:  Cannot parse .../syntax.py: SyntaxError: '(' was never closed (line 2)   exit 1
directory:    SKIP (import failed): Cannot parse .../malformed.py: SyntaxError: ...
              Generated CLI app: out/good_cli.py
```

Four regressions, since the risk here is over-correcting into noise:

- malformed file reports the parse error, single-file — and explicitly asserts the old "no app instance found" message is **gone**
- a valid non-app file still gets "no app instance found" and **not** "Cannot parse"
- a scan reports the bad file **and still converts the good one**
- a scan with an ordinary non-app module emits no `SKIP` at all

Two fail on the previous commit.

Also dropped the duplicated path in `discovery.py` — the exception already names the file, so it was printed twice.

## PR description

Both corrections applied, and the description was stale in more places than you listed once I reread it — it still described only three findings when there are now seven. Rewritten end to end: current counts, `load()` **raises** rather than exits, and each finding with its actual reproduction.

`make check`: 448 passed.

---

Four rounds, and the pattern is consistent enough to be worth naming: every finding has been in an `except` branch or an early-return that swallows something. #102 records that — error paths need their own probes, because the happy-path verification I keep running never reaches them. That guidance is what I should have been working from three rounds ago.

PR #113, opened 2026-08-26
fix: support declared FastMCP versions

tugrulguner:
## Problem

intpot declares `fastmcp>=2.0.0`, but inspection only understood FastMCP 3's `local_provider`. Real FastMCP 2 applications therefore appeared to contain zero tools. MCP parameter descriptions exposed by FastMCP's JSON Schema were also discarded, and the locked CI graph did not exercise the declared framework lower bounds.

## Approach

- Adapt structurally to FastMCP 2's `_tool_manager` and FastMCP 3's `local_provider` registries.
- Raise an actionable inspection error for unknown registry layouts and surface it cleanly through CLI commands.
- Read parameter descriptions from each tool's JSON Schema while unwrapping `Annotated` metadata to executable Python types.
- Add exact lower-bound lanes for Typer 0.9.0, FastAPI 0.100.0, and FastMCP 2.0.0, plus a latest FastMCP 2.x lane.
- Add a scheduled/manual newest-resolvable dependency lane.

## Verification

- `make check` — 458 tests passed; Ruff and Pyright passed.
- Actionlint 1.7.12 passed on `.github/workflows/ci.yml`.
- Typer 0.9.0 inspector suite — 6 passed.
- FastAPI 0.100.0 inspector suite — 10 passed.
- FastMCP 2.0.0 inspector suite — 9 passed.
- Latest FastMCP 2.x (2.14.7 at verification) inspector suite — 9 passed.
- End-to-end MCP-to-API regression compiles and executes generated source, verifies the OpenAPI parameter description, and invokes the endpoint through `TestClient`.

## Risk and scope

The adapters use structural attributes rather than version checks or private class identity. Unknown future layouts fail explicitly instead of silently losing tools. This does not redesign `ToolInfo` or broaden unrelated conversion semantics.

Closes #111

Issue #20, opened 2026-04-06
Dependency injection mapping — convert Depends() to CLI/MCP equivalents


tugrulguner:
## Summary

FastAPI's \`Depends()\` pattern is currently noted as a comment in converted code but not actually transformed. This issue tracks proper dependency injection mapping across frameworks.

## Example

```python
# FastAPI source
def get_db():
    db = SessionLocal()
    try:
        yield db
    finally:
        db.close()

@app.get("/users")
def list_users(db: Session = Depends(get_db)):
    return db.query(User).all()
```

Should convert to CLI as:
```python
@app.command()
def list_users():
    db = SessionLocal()
    try:
        users = db.query(User).all()
        typer.echo(users)
    finally:
        db.close()
```

## Challenges

- \`Depends()\` can be nested
- Generator-based deps need try/finally inlining
- Some deps are auth/middleware with no CLI equivalent
- Class-based dependencies

## Relevant files

- \`src/intpot/core/inspectors/api.py\` — already extracts dependency names
- \`src/intpot/core/transforms.py\` — transform infrastructure
- \`src/intpot/core/models.py\` — \`ToolInfo.dependencies\` field exists

PR #114, opened 2026-08-26
fix: reject unsupported FastAPI dependencies

### LINKED ISSUE #112: [Feature]: Refuse unsupported FastAPI dependency conversions
## What Happens Now

When converting a FastAPI endpoint to CLI or MCP, intpot removes `Depends(...)` parameters from the generated signature but preserves the original function body. If that body references the injected value, generated code can compile and then fail at invocation with `NameError`.

Only default-style `Depends(...)` is recorded today. FastAPI also normalizes `Annotated[..., Depends(...)]`, `Security(...)`, nested providers, and route-, router-, or application-level dependencies.

## Why I want it

intpot should never knowingly emit a program that cannot execute. Full dependency-provider mapping is tracked separately in #20; until that exists, unsupported conversions need to fail before writing broken output.

### AC

- [ ] Preserve FastAPI route inspection metadata, including normalized dependency information.
- [ ] Refuse FastAPI-to-CLI and FastAPI-to-MCP conversion when any normalized dependency is present.
- [ ] Detect default, `Annotated`, `Security`, nested, route-level, router-level, and application-level dependencies.
- [ ] Report an actionable error through both Python and CLI surfaces without a traceback.
- [ ] Do not partially write directory conversion output when one source is unsupported.
- [ ] Continue converting dependency-free FastAPI routes.
- [ ] Keep full provider mapping in #20 open.

#### Open question

The current intermediate representation stores dependency names as strings. A future executable mapping will need richer provider metadata, but this guard should avoid prematurely designing that representation.

### Proposed Solution

Collect names from FastAPI's normalized route dependency tree at the inspector boundary. Add a dedicated unsupported-conversion error at the API-to-CLI/MCP boundary, validate all directory inputs before writing, and surface the error cleanly from commands.

### WHERE

- `src/intpot/core/inspectors/api.py`
- `src/intpot/converter.py` and shared conversion command path
- Inspector, Python API, CLI, and directory behavior tests
- Public limitation documentation and shipped agent guidance

### HOW

Use structural route/dependency attributes across supported FastAPI versions. Do not inline providers or rewrite function bodies in this scope.

### DEPENDENCIES

Related to #20, which remains the full dependency-mapping feature.

### Alternatives

Continue emitting warning comments. This is rejected because comments do not prevent runtime-broken generated code.

### Feature Type

Conversion correctness and safety.

### Scope

Fail-closed behavior only. No dependency execution, inlining, lifecycle conversion, or target-framework equivalent is included.

### Contribution

Implementation and tests will be contributed.

tugrulguner:
## Problem

FastAPI-to-CLI/MCP conversion removed dependency-injected parameters but preserved endpoint bodies that could reference them. Generated code could compile and then fail at invocation with `NameError`. The existing inspector also captured only a subset of FastAPI's normalized dependency forms.

## Approach

- Traverse FastAPI's normalized route dependency graph structurally.
- Preserve dependency names for inspection across `Depends`, `Annotated` dependencies, `Security`, nested providers, and route/router/application dependency lists.
- Refuse API-to-CLI/MCP conversion with public `UnsupportedFastAPIDependencyError` before emitting broken code.
- Plan complete directory conversions before writing any output.
- Keep intpot's public CLI and Typer app conversion usable with Typer 0.9 and current Click releases after integrating the compatibility CI changes from #113.
- Keep all six advanced conversion artifacts and isolate expected `Depends()` rejection in a dedicated runnable example.
- Update documentation, shipped skills, generated examples, and the executable demo to describe and prove the fail-closed behavior.

## Verification

- `make check` — 477 tests passed; Ruff and Pyright passed.
- FastAPI 0.100.0 inspector suite — 14 passed.
- Typer 0.9.0 with Click 8.3.1 and 8.5.0 public CLI/conversion suites — 14 passed in each environment, including `Annotated` parameters and child-declared sub-app names.
- A nested Typer 0.9 application was converted through the installed `intpot` command, then compiled, imported, and invoked successfully.
- Regenerated, compiled, imported, and invoked both advanced FastAPI conversion artifacts successfully.
- `bash scripts/demo.sh --no-save` completed successfully and verified all expected dependency rejections.
- Directory regression verifies that one unsupported application prevents every planned output from being written.

## Risk and scope

Inspection remains available; only conversion to targets that cannot preserve FastAPI dependency semantics is rejected. Full dependency-provider execution and lifecycle mapping remain tracked in #20.

Closes #112


tugrulguner (commented):
Reviewed exact head `c0c32134e045301ebb0587d14d5a141912325533` against `main` (`a2c1aaa7ba59b3caf1bcac2840a9307e76168c38`).

No actionable findings.

Verified:
- Read the complete diff and all changed files in context.
- `make check`: Ruff, formatting, Pyright, and all 459 tests passed.
- FastAPI 0.100.0 inspector lane: all 14 tests passed against the real lower-bound package.
- `bash scripts/demo.sh --no-save` completed successfully and exercised both dependency rejection paths plus atomic directory rejection.
- Focused coverage confirms default/`Annotated` `Depends`, `Security`, nested providers, callable providers, and app/router/route-level dependencies remain inspectable and reject CLI/MCP conversion.
- GitHub lint, typecheck, Python 3.11–3.14, changelog, and build checks are green.

The guard is placed before transformation/generation, the Python error is public, the CLI removes tracebacks, and directory output is fully planned before any write, satisfying #112 while leaving dependency-free routes convertible.

tests/test_inspectors/test_cli_inspector.py
```diff
@@ -147,3 +147,21 @@ def test_an_unrecognised_parameter_type_falls_back_to_str():
     tools = CLIInspector().inspect(_StubGroup({"pick": command}))
 
     assert tools[0].parameters[0].type_annotation == "str"
+
+
+def test_parameter_help_falls_back_to_callback_metadata():
```
tugrulguner, line 178: **High / blocking — the Typer 0.9 compatibility check is false-green on the public surface.** This stub-level fallback test passes, and the exact CI command reports 7 passing inspector tests, but that environment resolves `typer==0.9.0` with `click==8.5.0`, where both `intpot --version` and `intpot to mcp examples/advanced_cli.py` exit 1 before dispatch with `RuntimeError: Type not yet supported: intpot.core.models.Agent | None`. Even bypassing the intpot command and calling `CLIInspector().inspect()` on the shipped `advanced_cli.py` fails: Typer cannot construct its boolean dual flag (`TypeError: Secondary flag is not valid for non-boolean flag`), the inspector swallows that construction error, and then raises `AttributeError: 'Typer' object has no attribute 'help'`. An independent reproduction confirms the boolean-option inspection failure already occurs with Click 8.3.1 and succeeds with Click 8.2.1, so this is not limited to the newest 8.5 release. So the new fallback and changelog claim do not establish usable Typer 0.9/current-Click compatibility. Please add a real public-CLI/example E2E to the exact lower-bound lane and either make that combination work end-to-end or adjust the declared compatibility/claim.

tugrulguner (commented):
Re-reviewed the new exact head after the merge reconciliation and follow-up compatibility/example commits. Found one blocking compatibility gap; see the inline finding. The FastAPI dependency guard, atomic directory planning, regenerated advanced artifacts, and dependency example otherwise verified cleanly.

tests/test_inspectors/test_cli_inspector.py
tugrulguner: Addressed in `c64f780c21b00fd66a6693951e915f5673f77966`.

You were right: the exact Typer 0.9 lane only exercised inspector units and missed both public failures.

- intpot's own command annotations now build under Typer 0.9, including `--version` with current Click flag semantics.
- `CLIInspector` falls back to Typer's registered callbacks when Typer cannot construct its Click tree.
- That fallback recursively preserves registered sub-app groups with cycle protection, after focused re-review caught the initial root-only implementation.
- The exact CI lane now runs the public `intpot --version` and `intpot to mcp examples/advanced_cli.py` paths as well as inspector regressions.

Verification:
- Typer 0.9.0 + Click 8.3.1: 11 passed.
- Typer 0.9.0 + Click 8.5.0: 11 passed.
- `make check`: 474 passed; Ruff and Pyright passed.
- The installed CLI converted the reviewer's nested Typer case; the generated MCP module compiled, imported, and invoked `tasks_search` successfully.
- Demo, Actionlint, Zizmor, and whitespace checks passed.

tugrulguner (commented):
Re-reviewed the exact updated head. The new lower-bound public-CLI tests fix the previously reported startup and shipped boolean-option path, but the registered-callback fallback still emits invalid Python for Typer's supported `Annotated` parameter style; see the blocking inline finding. Everything else verified cleanly: `make check` passed all 474 tests plus Ruff/format/Pyright, the FastAPI 0.100 inspector lane passed 14 tests, the full demo passed, regenerated advanced API artifacts matched the checked-in files and compiled, and a generated CLI command executed successfully.

src/intpot/core/inspectors/cli.py
```diff
@@ -60,14 +65,96 @@ def inspect(self, app: Any) -> list[ToolInfo]:
...
+                if raw_default is inspect.Parameter.empty or raw_default is Ellipsis:
+                    default: Any = _SENTINEL
+                else:
+                    default = raw_default
+
+                help_text = getattr(metadata, "help", "")
+                parameters.append(
+                    ParameterInfo(
+                        name=param.name,
+                        type_annotation=python_type_name(
```
tugrulguner: **High / blocking — the compatibility fallback generates invalid code for Typer's `Annotated` parameter style.** In the declared Typer 0.9 lane with either Click 8.3.1 or 8.5.0, `get_group()` fails on the boolean option and sends the whole app through this fallback. For a valid callback such as `query: Annotated[str, typer.Argument(help="Search query")]`, `python_type_name()` stringifies the complete annotation, including the runtime metadata repr. `intpot to mcp app.py` then exits 0 but emits `query: Annotated[str, <typer.models.ArgumentInfo object at 0x...>]`, and `python -m py_compile` fails with `SyntaxError`. The same happens to `Annotated[bool, typer.Option(...)]`. Please unwrap `Annotated` to its base type in this fallback (and read help/default metadata from its metadata entries) and add this form to the real Typer 0.9 public conversion test so the compatibility claim covers both supported Typer declaration styles.

src/intpot/core/inspectors/cli.py
```diff
@@ -60,14 +65,96 @@ def inspect(self, app: Any) -> list[ToolInfo]:
...
+            )
+
+        groups = getattr(app, "registered_groups", None)
+        if not isinstance(groups, list):
+            return
+        for group in groups:
+            nested = getattr(group, "typer_instance", None)
+            if nested is None:
+                continue
+            group_name = getattr(group, "name", None)
```
tugrulguner: **Medium — the fallback silently drops nested app names supplied on the child `Typer`.** With Typer 0.9, the common `child = typer.Typer(name="admin"); app.add_typer(child)` form stores the effective name in `group.typer_instance.info.name`; `group.name` remains a falsey `DefaultPlaceholder`. Under Click 8.5 the new fallback therefore extracts `list_users`, while the normal Click 8.1.7 path extracts `admin_list_users`. Besides changing the public tool name based only on the installed Click version, two sub-apps with the same command can now collide. The new nested-fallback test misses this because its stub puts a literal name directly on the group. Please fall back to the nested app's `info.name` when the registered-group name is a placeholder, and cover the real `Typer(name=...); add_typer(child)` form.

src/intpot/core/inspectors/cli.py
tugrulguner: Addressed in `2e026edd87e63c2388a5967c2b4f821ddd1fedc2`.

The registered-callback fallback now unwraps `Annotated` to its base type and reads Typer argument/option help and defaults from the annotation metadata. It also structurally distinguishes Typer metadata from unrelated preceding `Annotated` metadata, after focused re-review caught that edge case.

The exact lower-bound public test now converts a real nested Typer app using both `Annotated[str, typer.Argument(...)]` and `Annotated[bool, typer.Option(...)]`, compiles the generated MCP source, and verifies the normalized types/defaults/help. The generated artifact was also imported and `admin_search("needle", include_done=True)` invoked successfully.

Verification: Typer 0.9 + Click 8.3.1: 14 passed; Typer 0.9 + Click 8.5.0: 14 passed; full `make check`: 477 passed.

src/intpot/core/inspectors/cli.py
tugrulguner: Addressed in `2e026edd87e63c2388a5967c2b4f821ddd1fedc2`.

Registered-group traversal now accepts `group.name` only when it is non-empty text; otherwise it falls back to `group.typer_instance.info.name`. This preserves `admin_list_users`/`admin_search` for `child = typer.Typer(name="admin"); app.add_typer(child)` instead of changing names with the installed Click version.

Coverage includes the structural placeholder case and a real public Typer 0.9 conversion using a child-declared name. The generated `admin_search` artifact compiled, imported, and invoked successfully under the compatibility reproduction.

tugrulguner (commented):
Re-reviewed the complete exact head against Intpot’s stated vision, ROADMAP.md, issue #112, and the deferred dependency-mapping scope in #20. The fail-closed architecture is aligned: inspection preserves normalized dependency metadata, API-to-CLI/MCP conversion refuses unsupported semantics before generation, directory writes are planned atomically, and the PR does not prematurely implement provider execution. I found one medium verification gap; see the inline finding.

scripts/demo.sh
```diff
@@ -95,14 +117,20 @@ run "Advanced MCP → API  (notes server)" \
     "uv run intpot to api examples/advanced_mcp.py" \
     "$OUTDIR/advanced_mcp_to_api.py"
 
-run "Advanced API → CLI  (user CRUD with Body, Depends, Optional, json)" \
+run "Advanced API → CLI  (user management)" \
```
tugrulguner, line 120: **Medium — the changed runnable examples are generated but never executed.** The demo reaches all conversion commands and validates the new rejection paths, but `run()` only prints or saves source. The new lower-bound public tests likewise stop at `compile()` and string assertions. That does not satisfy this repository’s explicit generated-artifact contract in `src/intpot/AGENTS.md`: compile, import, and invoke real behavior. Both advanced API outputs changed here, and this PR adds a Typer compatibility fallback capable of producing syntactically valid but behaviorally wrong code. Please make the executable verification durable—for example, invoke a generated CLI command and a generated MCP tool, including the Typer 0.9 fallback path—rather than relying on the one-time manual verification recorded in the PR body. The architecture is aligned; this is the remaining coverage gap before readiness.

tugrulguner: Fixed in #116. The checked-in generated CLI and MCP artifacts are now imported and invoked through their real framework boundaries, the dependency FastAPI example is requested with `TestClient`, and the Typer 0.9 compatibility lane now imports and invokes its generated MCP tool instead of stopping at `compile()`. Local verification passed with 478 tests, both Click 8.3.1/8.5.0 Typer-floor lanes passed, the full demo executed all three example boundaries, and every GitHub check on #116 is green.

PR #115, opened 2026-08-27
feat: add encoding and overwrite controls to IntpotApp.write

guyua9:
## Summary

- add explicit `encoding` control to `IntpotApp.write()` (defaulting to UTF-8)
- add `overwrite=False` protection for callers that must preserve an existing output file
- add focused tests for both behaviors

## Why

`IntpotApp.write()` is a public Python API for materializing generated source, but it previously delegated to `Path.write_text()` without exposing the encoding and always replaced an existing destination. Callers generating files in a known encoding or running a safety-conscious workflow had no way to express either requirement.

The default remains compatible for normal callers: generated files use UTF-8 and existing files are still replaceable unless `overwrite=False` is requested.

## Validation

- `make check` — 460 tests passed
- `uv run python -m pytest tests/test_converter.py -q` — 14 passed
- `uv run pyright src/ tests/test_converter.py` — 0 errors
- `uv run ruff check src/ tests/test_converter.py`
- `uv run ruff format --check src/ tests/test_converter.py`
- `git diff --check`

No issue is linked; this is a focused public-API usability and safety improvement.

## AI assistance

AI-assisted. The current `main` implementation, existing converter tests, repository contribution guidance, and current open PR list were reviewed before implementation.


tugrulguner:
Contributor-flow heads-up—no action is needed unless #117 merges before this pull request.

#117 changes unreleased notes from pull-request-number filenames to either an underlying issue number or a Towncrier-generated orphan. This pull request explicitly has no linked issue, so after refreshing from `main`, please preserve the current note text while replacing `changelog.d/115.changed.md` with an orphan created by:

```bash
uv run towncrier create +.changed.md
```

The code changes merge cleanly with the exact #117 head; this is only a post-merge changelog-flow transition.

src/intpot/converter.py
```diff
@@ -133,10 +145,12 @@ def write(self, path: str | Path, target: str | SourceType) -> Path:
         generators = {"cli": self.to_cli, "mcp": self.to_mcp, "api": self.to_api}
         if target not in generators:
             raise ValueError(f"Unknown target '{target}', expected: cli, mcp, api")
-        code = generators[target]()
         out = Path(path)
+        if out.exists() and not overwrite:
```
tugrulguner, line 246: **High / blocking — `overwrite=False` is not atomic and can still overwrite another writer's file.** There is a check-to-write window between this `exists()` call and `write_text(..., mode='w')`. I reproduced it by creating the destination immediately after this check returned false; `write()` then returned successfully and replaced that content with the generated source. This violates the safety guarantee for concurrent build/watch processes and also leaves symlink races. Please generate the code first, create the parent, and open the destination with exclusive creation (`open('x', encoding=...)`) when `overwrite=False`, letting the filesystem atomically raise `FileExistsError`; add a deterministic race/exclusive-create regression test.

src/intpot/converter.py
```diff
@@ -133,10 +145,12 @@ def write(self, path: str | Path, target: str | SourceType) -> Path:
...
+            raise FileExistsError(f"Output file already exists: {out}")
+        code = generators[target]()
         out.parent.mkdir(parents=True, exist_ok=True)
-        out.write_text(code)
+        out.write_text(code, encoding=encoding)
```
tugrulguner, line 250: **High / blocking — the advertised encoding example can produce a `.py` file that Python cannot load.** The new test writes with `encoding='utf-16'` and only reads the text back. Against this exact head, writing even `print("generated")` as UTF-16 succeeds, but `python -m py_compile generated.py` fails with `SyntaxError: source code string cannot contain null bytes`. Other non-UTF-8 encodings require a valid PEP 263 declaration when generated source contains non-ASCII text, while UTF-16 is not suitable for Python source because the declaration itself must be ASCII-compatible. Please define and enforce the supported source encodings, emit an appropriate coding declaration where needed, reject unsupported encodings such as UTF-16, and make the test compile/import the written artifact rather than only decoding it.

tugrulguner (changes requested):
Reviewed exact head `a309fdbb67d273417d3004db6e970e29a314a732`. The API shape is clear and the default remains backward-compatible, but two guarantees added by this PR are not currently upheld: the documented UTF-16 example writes a `.py` file Python cannot compile, and `overwrite=False` can still overwrite a file created between the existence check and the write. See the inline blockers.

CI is failing only in the changelog job: `changelog.d/115.changed.md` uses the pull-request number, while numeric fragment names must reference an underlying issue. Because this PR explicitly has no issue, replace it with a unique orphan fragment generated via `uv run towncrier create +.changed.md` (keeping the existing user-facing sentence in the generated file).

Verification: `make check` passed Ruff, formatting, Pyright, and all 488 tests; focused `tests/test_converter.py` passed all 17 tests; the two edge cases above were reproduced separately against the exact head. Please also update the README/Python skill example to expose the new public `encoding` and `overwrite` options once their semantics are corrected.

tugrulguner:
## CI failure

The code-quality and test jobs are passing. The failing `changelog` job reports:

```text
changelog.d/115.changed.md must use an issue number, not a pull request number
```

Numeric changelog fragment names refer to the underlying issue, never the PR. This PR states that it has no linked issue, so `115.changed.md` is invalid because `#115` is this pull request.

Please replace it with a unique orphan fragment:

```bash
uv run towncrier create +.changed.md
```

Keep the current user-facing sentence in the newly generated `changelog.d/+<unique-id>.changed.md` file and remove `changelog.d/115.changed.md`. That will satisfy the changelog workflow; `skip-changelog` is not appropriate because this is user-facing.

tugrulguner (changes requested):
Re-reviewed exact head `914a67d7c0c90ccd9f151ab1ac2dc1d1d409640e`. The only update since the prior review is a merge from `main`; the PR implementation and its two blocking defects are unchanged:

- `overwrite=False` still performs a racy `exists()` check before a normal truncating write. I reproduced a file created in that window being overwritten. Please use atomic exclusive creation for the non-overwrite path and add a deterministic regression test. Prior discussion: https://github.com/tugrulguner/intpot/pull/115#discussion_r3880494924
- The test still advertises UTF-16 for generated Python source without compiling the artifact. Exact-head reproduction fails with `PyCompileError: SyntaxError: source code string cannot contain null bytes`. Define/enforce supported source encodings and compile/import the generated file in the test. Prior discussion: https://github.com/tugrulguner/intpot/pull/115#discussion_r3880494929

Two repository requirements also remain unresolved:

- The changelog job fails because `changelog.d/115.changed.md` uses this pull request number despite there being no underlying issue. Replace it with a Towncrier-generated orphan fragment (`uv run towncrier create +.changed.md`) while preserving the sentence.
- This changes the public `IntpotApp.write()` API, so update both the README examples/reference and `src/intpot/templates/skills/intpot-python.md` to document `encoding` and `overwrite`, per `AGENTS.md`.

Local verification after `uv sync --all-extras`: `make check` passed Ruff, formatting, Pyright, and all 488 tests. CI code/test jobs pass; the changelog job is still failing.

PR #125, opened 2026-08-31
feat: add canonical application schema

tugrulguner:
## Summary
- add immutable `ApplicationSchema`, `ToolSchema`, and `ParameterSchema` as Intpot's canonical semantic core
- route runtime registration, inspection, target projection, and generation through the same schema while preserving detached `ToolInfo` compatibility
- preserve application names across generated CLI, API, and MCP targets and regenerate executable conversion examples
- document the public schema and projection APIs for the v0.8 foundation

## Verification
- `make check` (497 passed; Ruff, format, and Pyright clean)
- `uv run python scripts/verify_generated_examples.py`
- built wheel and sdist, installed each in separate Python 3.11 environments outside the checkout
- generated a FastAPI app through the installed wheel and invoked `/add` with `TestClient` (`2 + 5 = 7`)
- verified wheel metadata/templates and sdist `pyproject.toml`/README contents


src/intpot/core/models.py
```diff
@@ -49,6 +53,88 @@ def __repr__(self) -> str:
...
+                for key, item in value.items()
+            )
+        )
+    if isinstance(value, list):
+        return _FrozenList(_freeze_default(item) for item in value)
+    if isinstance(value, tuple):
+        return _FrozenTuple(_freeze_default(item) for item in value)
+    if isinstance(value, set):
+        return _FrozenSet(_freeze_default(item) for item in value)
+    return copy.deepcopy(value)
```
tugrulguner: **High — the canonical schema is not actually deeply immutable.** The fallback only deep-copies unknown values and then exposes them directly. A standard mutable default is enough to mutate the frozen schema:

```python
p = ParameterSchema(name="payload", default=bytearray(b"a"))
p.default[0] = ord("z")
assert p.default == bytearray(b"z")
```

User-defined mutable/hashable defaults are worse: changing the nested object changes the generated frozen-dataclass hash and can corrupt a set/dict containing the schema. Please freeze the remaining supported mutable values (for example `bytearray`/`deque`) and reject or convert opaque mutable defaults rather than treating `deepcopy` as immutability. Add an adversarial test that mutates through `ParameterSchema.default`.

src/intpot/core/models.py
```diff
@@ -158,3 +244,169 @@ def __post_init__(self) -> None:
...
+        """Return a sentinel-free representation suitable for JSON encoding."""
+        result: dict[str, Any] = {
+            "name": self.name,
+            "type_annotation": self.type_annotation,
+            "description": self.description,
+            "param_source": self.param_source.value if self.param_source else None,
+            "required": self.required,
+        }
+        if not self.required:
+            result["default"] = _thaw_default(self.default)
```
tugrulguner: **Medium — `to_dict()` does not satisfy its JSON-compatible contract.** `_thaw_default()` restores sets as sets, so a value explicitly handled by `_freeze_default()` still cannot be encoded:

```python
data = ParameterSchema(name="tags", default={"a", "b"}).to_dict()
json.dumps(data)  # TypeError: Object of type set is not JSON serializable
```

`Path`, enum, bytes, and opaque defaults have the same issue. Please use a dedicated JSON-normalization path (with a defined representation for sets and non-JSON scalars) rather than the generation-oriented thaw path, and test `json.dumps(schema.to_dict())`.

src/intpot/runtime.py
```diff
@@ -69,14 +76,24 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
                 func, name_override=name, description_override=description
             )
             self._tools.append(RegisteredTool(func=func, info=info))
+            self.__dict__.pop("schema", None)
             return func
 
         return decorator
 
+    @functools.cached_property
```
tugrulguner, line 127: **Medium — caching makes the generated app name depend on whether `.schema` was accessed before a public `App.name` update.** Reproduction:

```python
app = App("first")
app.tool()(lambda: "ok")
_ = app.schema
app.name = "second"

assert build_fastapi_app(app.name, app._tools).title == "second"
assert "title='first'" in app.eject("api")
```

That makes live serving and ejection disagree, and the result changes if the first schema access is removed. Either make `name` immutable, invalidate the cached schema when it changes, or avoid caching the runtime schema; add a serve/eject name-parity test after a rename.

tugrulguner (commented):
## Review

Changes requested: one high-severity contract bug and two medium inconsistencies. The overall schema/projection direction is coherent and the full matrix is green, but the headline immutability and serialization guarantees are not yet true for supported Python values.

### Verification
- `make check`: 497 passed; Ruff, format, and Pyright clean
- latest dependency overlay: 497 passed
- generated example verifier: CLI, MCP, and FastAPI behavior passed
- wheel and sdist built successfully
- exact-head GitHub checks are green

Reviewed head: `2b55c3864c8412f3df1a8a0e90d4d069a23cbf9d`

src/intpot/converter.py
```diff
@@ -47,6 +47,50 @@ def inspect_app(source_type: SourceType, app_instance: Any) -> list[ToolInfo]:
     return APIInspector().inspect(app_instance)
```
tugrulguner, line 53: **Medium — the newly public `compile_app()` cannot compile the newly public `SourceType.PYTHON`.** `_inspect_tools()` handles MCP and CLI and routes every other enum value to `APIInspector`, so the apparent public combination fails:

```python
app = intpot.App("demo")
app.tool()(lambda x: x)
intpot.compile_app(intpot.SourceType.PYTHON, app)
# AttributeError: 'App' object has no attribute 'routes'
```

Please add an explicit Python branch that compiles `intpot.App` registrations (or returns its canonical schema), or reject/document this combination with a deliberate `ValueError` instead of accidentally treating Python as FastAPI. Add coverage for every public `SourceType` value accepted by `compile_app()`.

tugrulguner (commented):
Follow-up on the same reviewed head: one additional medium public-API inconsistency was confirmed by the parallel compatibility pass.

src/intpot/core/models.py
tugrulguner: Fixed in `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`. `bytearray` and `deque` defaults now become immutable snapshots, dictionary keys are frozen recursively, and unsupported opaque defaults fail deliberately with `TypeError`. The adversarial mutation/hashable-object regression is in `test_parameter_schema_freezes_mutable_defaults_and_rejects_opaque`.

src/intpot/core/models.py
tugrulguner: Fixed in `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`. `to_dict()` now uses a dedicated deterministic JSON-normalization path with tagged representations for sets, bytes/bytearrays, paths, enums, deques, and other supported non-JSON scalars. `test_schema_to_dict_normalizes_supported_non_json_defaults` executes `json.dumps(schema.to_dict())`.

src/intpot/runtime.py
tugrulguner: Fixed in `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`. `App.name` is now a property whose setter invalidates the cached schema, preserving stable snapshots while keeping live serving and ejection consistent after a public rename. `test_runtime_schema_and_ejection_follow_public_app_renames` exercises both paths.

src/intpot/converter.py
tugrulguner: Fixed in `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`. `compile_app(SourceType.PYTHON, app)` now has an explicit `intpot.App` inspection path and preserves the runtime app name; non-App inputs receive a deliberate `ValueError`. Covered by `test_compile_app_supports_the_public_python_source_type`.

tugrulguner:
Addressed all four review findings in `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`:

- deep-froze supported mutable defaults and rejected opaque mutable defaults
- added deterministic JSON normalization for supported non-JSON defaults
- invalidated the cached runtime schema when `App.name` changes
- added explicit `compile_app(SourceType.PYTHON, App)` support

Verification on the final tree: `make check` passed; generated CLI/MCP/FastAPI examples executed; the final wheel and sdist were each installed in isolated Python 3.11 environments and smoke-tested; exact-head CI passed across Python 3.11–3.14 and all compatibility lanes. All four inline threads have replies and are resolved.

tugrulguner (commented):
Exact-head re-review completed for `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`. The diff from reviewed head `2b55c3864c8412f3df1a8a0e90d4d069a23cbf9d` addresses all four findings with regression coverage. Local gates, final distribution artifacts, and exact-head CI are green. No remaining findings.

src/intpot/core/models.py
```diff
@@ -49,6 +60,259 @@ def __repr__(self) -> str:
...
+        return _FrozenList(_freeze_default(item) for item in value)
+    if isinstance(value, tuple):
+        return _FrozenTuple(_freeze_default(item) for item in value)
+    if isinstance(value, set):
+        return _FrozenSet(_freeze_default(item) for item in value)
+    if isinstance(value, frozenset):
+        return frozenset(_freeze_default(item) for item in value)
+    if value is None or value is Ellipsis or value is NotImplemented:
+        return value
+    if isinstance(
```
tugrulguner: **High — supported scalar subclasses and enum defaults still bypass deep freezing.** This `isinstance` branch deep-copies and then exposes the object directly. A mutable `str` subclass is accepted, and mutating an attribute after construction changes `hash(ParameterSchema)`, recreating the set/dict-corruption problem this fix intended to close. Standard enums can also carry mutable values: `class Values(Enum): item = []`; after `p = ParameterSchema(name="v", default=Values.item)`, `p.default.value.append("changed")` mutates `p.to_dict()`. A dict-valued enum additionally makes `to_dict()` raise `TypeError` because the raw enum payload never passed through `_freeze_default`. Please normalize exact built-in immutable values rather than retaining mutable subclasses, and recursively snapshot or deliberately reject enum payloads; add regressions for hash stability, enum mutation, and `json.dumps()` of supported enum values.

tugrulguner (commented):
## Review

Changes requested: one high-severity immutability/serialization gap remains. The overall semantic-core integration is coherent, all prior threads are resolved, and the repository and generated-example gates are green, but the canonical schema still exposes mutable state through values explicitly accepted as supported.

### Verification
- `make check`: 501 passed; Ruff, format, and Pyright clean
- generated example verifier: CLI, MCP, and FastAPI behavior passed
- exact-head GitHub checks are green

Reviewed head: `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec`

src/intpot/core/models.py
```diff
@@ -49,6 +60,259 @@ def __repr__(self) -> str:
...
+            date,
+            datetime,
+            time,
+            timedelta,
+            Decimal,
+            Enum,
+            Fraction,
+            Path,
+            range,
+            slice,
```
tugrulguner: **High — `slice` is explicitly accepted without freezing its components.** Slice bounds can be arbitrary mutable objects: `p = ParameterSchema(name="window", default=slice([], None)); p.default.start.append("changed")` mutates the supposedly immutable schema and changes `p.to_dict()`. On supported Python 3.11, even `hash(ParameterSchema(name="window", default=slice(1, 2)))` raises `TypeError` because `slice` is unhashable. Please snapshot each bound recursively into an immutable representation (or reject slices deliberately) and cover both mutation and Python 3.11 hash behavior.

src/intpot/runtime.py
```diff
@@ -69,14 +85,24 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
...
+    def schema(self) -> ApplicationSchema:
+        """Return a stable snapshot, invalidated by registrations or renaming."""
+        return ApplicationSchema.from_tools(
+            name=self.name,
+            source_type=SourceType.PYTHON,
+            tools=(tool.info for tool in self._tools),
+        )
+
     @property
     def tools(self) -> list[ToolInfo]:
```
tugrulguner, line 137: **High — the documented compatibility `.tools` edge now fails for defaults it previously exposed.** Before this PR, `App.tools` returned the registered `ToolInfo` records directly. Now it compiles through `ApplicationSchema`, so a valid registration such as `class Marker: pass; marker = Marker(); @app.tool() def use_marker(value=marker): ...` makes `app.tools` raise `TypeError: Unsupported parameter default ...`. This breaks the compatibility API before callers even request generation. Preserve detached `ToolInfo` access for legacy/default values independently of schema compilation, or provide a compatibility representation that does not retain mutable state inside the canonical schema.

changelog.d/+semantic-core.added.md
```diff
@@ -0,0 +1 @@
+Intpot now compiles runtime and converted applications into an immutable, inspectable `ApplicationSchema` with target projections shared by generation.
```
tugrulguner, line 1: **Medium — the fragment records only an addition, but this PR also changes public contracts.** `App.tools` changes from live registered records to detached snapshots (and currently rejects opaque defaults), while `intpot inspect --json` changes supported non-JSON defaults from `default=str` strings to tagged objects. Add a `changed` fragment and migration note, or preserve the prior contracts; an `added` fragment alone makes the 0.8 release notes incomplete.

tugrulguner (commented):
Follow-up compatibility pass on the same exact head found two additional blocking contract gaps and one release-note gap.

src/intpot/core/models.py
tugrulguner: **High — values accepted as supported defaults are not necessarily renderable by any generator.** All three templates emit `repr(default)`. A normal `Path("x")` default becomes `PosixPath('x')`, while the preserved import is typically only `from pathlib import Path`, so importing the ejected artifact raises `NameError`. An enum default renders as `<Color.red: 'red'>`, which is a `SyntaxError`. The canonical schema is now the direct generation contract, so its accepted-default domain needs a source-literal renderer plus required imports (or it must reject values it cannot render). Add real import/execution tests for every supported structured default category.

src/intpot/core/models.py
```diff
@@ -49,6 +60,259 @@ def __repr__(self) -> str:
...
+
+
+class _FrozenSet(frozenset[Any]):
+    """Immutable set value that renders as a normal set literal."""
+
+    def __repr__(self) -> str:
+        return repr({_thaw_default(value) for value in self})
+
+
+class _FrozenBytearray(bytes):
```
tugrulguner, line 121: **Medium — frozen wrappers collapse semantically different default types in equality and hashing.** `_FrozenBytearray` subclasses `bytes`, `_FrozenList` subclasses `tuple`, and `_FrozenSet` subclasses `frozenset`, so `ParameterSchema("x", default=bytearray(b"a")) == ParameterSchema("x", default=b"a")` even though `to_info()` restores different types and `to_dict()` emits different tagged representations. The same occurs for list/tuple and set/frozenset. Schema equality should distinguish values that thaw and serialize differently; use type-tagged composition or type-sensitive equality/hash behavior.

src/intpot/core/models.py
```diff
@@ -49,6 +60,259 @@ def __repr__(self) -> str:
...
+    if isinstance(value, _FrozenTuple):
+        return tuple(_thaw_default(item) for item in value)
+    if isinstance(value, _FrozenSet):
+        return {_thaw_default(item) for item in value}
+    if isinstance(value, frozenset):
+        return frozenset(_thaw_default(item) for item in value)
+    return copy.deepcopy(value)
+
+
+def _json_key(value: Any) -> str | int | float | bool | None:
```
tugrulguner: **Medium — JSON normalization can silently lose distinct mapping keys.** For `default={1: "integer", "1": "string"}`, `to_dict()` retains two Python keys but `json.loads(json.dumps(...))` has only one because JSON stringifies both as `"1"`. Structured keys can likewise collide with strings containing their encoded representation. Since this is documented as a transparent JSON view, encode mappings with non-string keys as tagged key/value entries or reject collisions; add a mixed-key round-trip regression.

src/intpot/runtime.py
```diff
@@ -101,7 +127,7 @@ def eject(self, target: str) -> str:
 
         mod = importlib.import_module(module_path)
         generator_cls = getattr(mod, class_name)
-        return generator_cls().generate(self.tools)
+        return generator_cls().generate(self.schema)
```
tugrulguner, line 167: **Medium — normal schema-validation failures escape the user-facing CLI as full tracebacks.** With an opaque default such as `value=object()`, `intpot eject app.py --to cli` reaches this schema compile and raises `TypeError`; `commands/eject.py` catches only `ValueError`. The same compile boundary is used by conversion/inspection helpers that catch only inspection/dependency errors. If unsupported defaults are an intentional validation error, map them to a concise stderr message and controlled nonzero exit consistently across CLI commands, with no traceback.

tugrulguner (commented):
Final generated-artifact and serialization pass on the same exact head found one additional high-severity generation gap and three medium contract/error-path gaps.

tugrulguner:
## Consolidated review status

**Changes requested:** 8 unresolved contract issues remain on exact head `3f758f0ca18eb4a1a46fe55a0f76035dd33dc4ec` (4 high, 4 medium).

### High
- `ParameterSchema` retains mutable scalar subclasses and enum payloads, so the advertised immutable schema can still expose mutable state.
- `slice` defaults retain mutable bounds and are unhashable on supported Python 3.11.
- Compatibility `.tools` now fails for opaque defaults that worked before this change.
- Accepted `Path` and enum defaults generate modules that fail to import (`NameError` / `SyntaxError`).

### Medium
- The changelog omits the `.tools` and `inspect --json` contract changes.
- Frozen list/tuple, set/frozenset, and bytearray/bytes defaults collapse in equality despite thawing and serializing differently.
- JSON normalization silently collapses distinct mixed-type mapping keys.
- Unsupported-default validation escapes CLI commands as a traceback instead of a controlled user-facing error.

Each finding has a reproduction and requested regression coverage in its inline thread. The four earlier findings are fixed and resolved.

### Verification
- Exact-head GitHub checks are green.
- Focused schema/eject suite: 21 passed on Python 3.14.
- Adversarial exact-head run independently reproduced slice mutation, equality collapse, mixed-key JSON loss, `.tools` compatibility failure, and invalid generated `Path`/enum modules.

src/intpot/core/models.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Scalar subclasses are normalized to exact immutable built-ins, while enum defaults are deliberately rejected because standalone generation cannot reliably preserve their defining class. Regression coverage verifies hash stability and mutable enum rejection.

src/intpot/core/models.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Slice bounds are now recursively frozen in a hashable `_FrozenSlice` snapshot and thawed only for compatibility access. The regression mutates the original bound and verifies both snapshot stability and hashing on Python 3.11.

src/intpot/runtime.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. `App.tools` now deep-copies registered `ToolInfo` directly, independent of canonical-schema compilation and cache state. Opaque defaults remain available through the detached compatibility edge, while `App.schema` deliberately rejects them.

changelog.d/+semantic-core.added.md
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Added a `changed` fragment documenting stable detached `App.tools`, tagged strict-JSON normalization, unsupported enum/opaque defaults, and executable source rendering. `towncrier build --draft --version 0.8.0` now includes both Added and Changed sections.

src/intpot/core/models.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Generators now use a dedicated recursive source-default renderer and collect required imports for every supported structured category. Path defaults use qualified `pathlib.Path` to avoid colliding with FastAPI's `Path`. All three targets are compiled/imported in regression coverage; wheel and sdist installed-artifact smokes also imported generated FastAPI source.

src/intpot/core/models.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Frozen list/tuple, set/frozenset, and bytearray/bytes representations now use type-sensitive equality and stable hashing. Bidirectional schema comparisons retain distinct source container semantics.

src/intpot/core/models.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Mappings with any non-string key now serialize as tagged ordered key/value entries, preventing JSON key coercion and mixed-key data loss. The regression round-trips `{1: 'integer', '1': 'string'}` through `json.dumps`/`json.loads`.

src/intpot/runtime.py
tugrulguner: Fixed in `c09b7d3b55a9cb5827a2ec947923555c9205527c`. Eject, inspect, and conversion command boundaries now map schema-validation `TypeError`s to concise stderr plus controlled exit 1. Regressions cover all three commands and assert no traceback.

tugrulguner (commented):
Exact-head re-review completed for `c09b7d3b55a9cb5827a2ec947923555c9205527c`. All eight unresolved schema-contract findings and the three follow-up adversarial findings are addressed with regression coverage. Local `make check` passed with 515 tests; generated CLI/MCP/FastAPI examples passed; wheel and sdist installed-artifact FastAPI generation/import smokes passed; exact-head CI is green across Python 3.11–3.14 and compatibility lanes. No remaining findings.

tugrulguner:
Addressed the remaining semantic-core review findings in `c09b7d3b55a9cb5827a2ec947923555c9205527c`:

- deeply freezes and type-tags supported defaults while rejecting enums/opaque values deliberately
- renders supported defaults as executable source with collision-safe imports
- preserves stable detached `App.tools` compatibility access
- emits strict JSON for mixed mapping keys and non-finite numbers
- converts schema-validation failures into concise CLI errors across eject, inspect, and conversion

Verification: `make check` (515 passed), generated-example verifier, wheel and sdist isolated installs plus generated FastAPI import smoke, exact-head CI green. All review threads are replied to and resolved; exact-head re-review found no remaining issues.

src/intpot/core/generators/_render.py
```diff
@@ -102,6 +104,13 @@ def _collect_extra_imports(tools: list[ToolInfo]) -> list[str]:
             if any(marker in imp for marker in _FRAMEWORK_IMPORT_MARKERS):
                 continue
             result.append(imp)
+        for parameter in tool.parameters:
+            if parameter.required:
+                continue
+            for imp in sorted(_default_imports(parameter.default)):
```
tugrulguner: **High — the declared legacy `Sequence[ToolInfo]` generator input still emits invalid source for structured defaults.** `GenerationInput` explicitly retains the compatibility sequence, but `_default_imports()` and `_source_default()` recurse only through frozen schema wrappers. For a raw `ToolInfo` with `default=deque([1])`, `APIGenerator().generate([tool])` emits `deque([1])` without importing `deque`; with `default=[Path('y')]` it emits `[PosixPath('y')]` and only adds `import pathlib`. Importing the generated module raises `NameError`. Either compile compatibility inputs through `ApplicationSchema` before rendering or make source/import traversal handle raw supported containers recursively. Add compile/import tests using `Sequence[ToolInfo]` for all three generators.

src/intpot/runtime.py
```diff
@@ -69,14 +86,24 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
...
+            name=self.name,
+            source_type=SourceType.PYTHON,
+            tools=(tool.info for tool in self._tools),
+        )
+
     @property
     def tools(self) -> list[ToolInfo]:
-        """Return ToolInfo list for all registered tools."""
-        return [t.info for t in self._tools]
+        """Return detached compatibility models without requiring schema support."""
```
tugrulguner: **High — the opaque-default compatibility fix is still incomplete.** `copy.deepcopy()` makes a normal user class work, but formerly valid opaque defaults need not be deepcopyable: `threading.Lock()` makes `app.tools` raise `TypeError: cannot pickle '_thread.lock' object`. A custom `__deepcopy__` returning `self` also makes the supposedly detached result alias the registered default, so mutating `app.tools[0].parameters[0].default` mutates runtime state. The resolved thread and changelog now promise stable detached compatibility access. Please handle non-copyable/aliasing defaults deliberately (or narrow and document the contract without claiming compatibility), with regressions for both cases.

src/intpot/core/models.py
```diff
@@ -49,6 +61,454 @@ def __repr__(self) -> str:
...
+    if isinstance(value, float):
+        return float(value)
+    if isinstance(value, complex):
+        return complex(value)
+    if isinstance(value, str):
+        return str(value)
+    if isinstance(value, bytes):
+        return bytes(value)
+    if isinstance(value, datetime):
+        return datetime.fromisoformat(value.isoformat())
```
tugrulguner, line 255: **Medium — exact built-in normalization silently loses `datetime`/`time` semantics.** The `isoformat()` round-trip drops `fold`, and replaces named/custom timezone objects with a fixed offset (or drops them for `time` values whose offset cannot be resolved without a date). For example, `ParameterSchema('x', default=datetime(2026, 11, 1, 1, 30, fold=1)).to_info().default.fold` is `0`; the same happens for `time(..., fold=1)`. Preserve the supported value's fold/timezone semantics in the frozen/source representation, or reject cases that standalone generation cannot preserve. Tests comparing datetimes do not catch this because equality ignores `fold`.

src/intpot/core/models.py
```diff
@@ -49,6 +61,454 @@ def __repr__(self) -> str:
...
+        if all(type(key) is str for key, _ in value.entries):
+            return {key: _json_default(item) for key, item in value.entries}
+        return {
+            "type": "mapping",
+            "items": [
+                {"key": _json_default(key), "value": _json_default(item)}
+                for key, item in value.entries
+            ],
+        }
+    if isinstance(value, (_FrozenList, _FrozenTuple, list, tuple)):
```
tugrulguner: **Medium — the documented transparent JSON view still conflates distinct defaults.** A list and tuple both normalize to the same JSON array, and a real bytes default collides with an ordinary mapping that happens to use the tag shape: `ParameterSchema('x', default=b'abc').to_dict()['default'] == ParameterSchema('x', default={'type': 'bytes', 'base64': 'YWJj'}).to_dict()['default']`. This defeats inspection of the source semantics and makes the tag namespace ambiguous. Use an unambiguous envelope/tagging strategy (including source container type where it matters), and add non-collision regressions for ordinary mappings versus every tagged representation.

src/intpot/core/models.py
```diff
@@ -49,6 +61,454 @@ def __repr__(self) -> str:
...
+
+    def __iter__(self) -> Iterator[Any]:
+        return (key for key, _ in self.entries)
+
+    def __len__(self) -> int:
+        return len(self.entries)
+
+    def __getitem__(self, key: Any) -> Any:
+        for candidate, value in self.entries:
+            if candidate == key:
```
tugrulguner: **Medium — `_FrozenDict` cannot retrieve an accepted NaN key that it exposes during iteration.** With `p = ParameterSchema('x', default={float('nan'): 'value'})`, `key = next(iter(p.default))` followed by `p.default[key]` raises `KeyError` because even the identical NaN object does not equal itself. Consequently `dict(p.default)` also fails, violating the public default's `Mapping` behavior. Match dict lookup's identity-or-equality semantics (`candidate is key or candidate == key`) and add a regression.

src/intpot/core/models.py
```diff
@@ -49,6 +61,454 @@ def __repr__(self) -> str:
...
+    if isinstance(value, time):
+        return time.fromisoformat(value.isoformat())
+    if isinstance(value, timedelta):
+        return timedelta(
+            days=value.days,
+            seconds=value.seconds,
+            microseconds=value.microseconds,
+        )
+    if isinstance(value, Decimal):
+        return Decimal(value)
```
tugrulguner, line 275: **Medium — the accepted Decimal domain still breaks frozen-schema hashing.** `ParameterSchema('x', default=Decimal('sNaN'))` constructs and serializes successfully, but `hash(parameter)` raises `TypeError: Cannot hash a signaling NaN value`. This is the same hashability contract the slice fix now tests. Reject signaling NaNs or store a stable immutable Decimal representation, and cover hash/equality behavior for special Decimal values.

tugrulguner (commented):
## Exact-head fix re-review

The eight previously unresolved threads are addressed, but adversarial compatibility and value-domain testing found two high-severity gaps and four medium contract gaps. See inline comments.

Verification: `make check` passed (515 tests; Ruff, format, and Pyright clean), generated CLI/MCP/FastAPI examples executed successfully, the changelog draft contains Added and Changed sections, and exact-head CI is green.

Reviewed head: `c09b7d3b55a9cb5827a2ec947923555c9205527c`

tugrulguner (commented):
Follow-up independent compatibility pass on the same exact head found two additional high-severity generation collisions and two medium contract/documentation gaps. These are separate from the six findings in the preceding review.

src/intpot/core/generators/_render.py
```diff
@@ -102,6 +104,13 @@ def _collect_extra_imports(tools: list[ToolInfo]) -> list[str]:
             if any(marker in imp for marker in _FRAMEWORK_IMPORT_MARKERS):
```
tugrulguner, line 147: **High — preserved `pathlib.Path` imports can still overwrite FastAPI's parameter-source `Path`.** A tool with `source_imports=('from pathlib import Path',)` and a `ParamSource.path` parameter generates both `from fastapi import FastAPI, Body, Path` and `from pathlib import Path`. The latter wins, so module import evaluates `Path(...)` as `pathlib.Path(...)` and raises `TypeError: expected str, bytes or os.PathLike object, not ellipsis`. The default renderer avoided this collision with qualified `pathlib.Path`, but preserved source imports bypass that protection. Detect/alias conflicting imports or qualify FastAPI parameter factories, and add an import test combining a path-sourced parameter with a body that uses `pathlib.Path`.

src/intpot/core/models.py
```diff
@@ -49,6 +61,454 @@ def __repr__(self) -> str:
...
+        return (
+            f"slice({_source_default(value.start)}, {_source_default(value.stop)}, "
+            f"{_source_default(value.step)})"
+        )
+    if isinstance(value, float) and not math.isfinite(value):
+        return f"float({str(value)!r})"
+    if isinstance(value, complex):
+        return f"complex({_source_default(value.real)}, {_source_default(value.imag)})"
+    if isinstance(value, datetime):
+        return f"datetime.fromisoformat({value.isoformat()!r})"
```
tugrulguner: **High — generated default imports are still vulnerable to tool-name shadowing.** If an earlier generated tool is named `datetime`, the later expression `datetime.fromisoformat(...)` resolves to that function, not the imported class, and module import raises `AttributeError: 'function' object has no attribute 'fromisoformat'`. The same collision applies to `date`, `time`, `timedelta`, `Decimal`, `Fraction`, `UUID`, `deque`, and even `pathlib`. Use collision-proof qualified/private aliases for generated-default dependencies and test multi-tool modules where tool names match every helper import. The changelog's current “without import collisions” claim is not yet true.

src/intpot/core/models.py
```diff
@@ -158,3 +618,169 @@ def __post_init__(self) -> None:
...
+
+    name: str
+    type_annotation: str = "str"
+    default: Any = _SENTINEL
+    description: str = ""
+    param_source: ParamSource | None = None
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "name", sanitize_identifier(self.name))
+        object.__setattr__(self, "default", _freeze_default(self.default))
```
tugrulguner, line 757: **Medium — schema equality still collapses distinct scalar source types.** Exact scalar normalization preserves `bool` and `int` separately, but dataclass equality delegates to Python value equality, so `ParameterSchema('x', default=True) == ParameterSchema('x', default=1)` and their hashes are identical even though `to_info()` restores different types. The fix made container equality type-sensitive for this reason; scalar defaults need the same source-type identity, either through tagged frozen values or an explicit equality/hash key. Add bidirectional regressions for `bool`/`int` and other cross-type equalities.

README.md
```diff
@@ -327,10 +336,16 @@ produces it; everything downstream consumes it.
...
-or `eject` sends it through the generators.
+Conversion follows four stages: detect the framework, inspect its tools, compile one
+immutable `ApplicationSchema`, then project target-specific semantics before rendering.
+The runtime side compiles `@app.tool()` registrations into the same schema. `serve` keeps
+the registered callables for live execution; `eject` renders detached views of the stable
+schema. Access `.schema.to_dict()` when a human, build tool, or agent needs to inspect
+exactly what Intpot compiled.
+
+`to_dict()` is directly JSON-serializable. Supported non-JSON defaults such as sets,
+bytes, paths, and enums use tagged dictionaries with a `type` field. Opaque mutable
```
tugrulguner: **Medium — public documentation still advertises enum serialization after enums were changed to validation errors.** This says enums use tagged dictionaries, while `_freeze_default()` now rejects every `Enum` with `TypeError` and the new changelog documents that rejection. Remove enums from the supported-default list and state the restriction consistently in the README and generated Python skill.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. Compatibility `Sequence[ToolInfo]` inputs now compile through `ToolSchema.from_info()` before rendering, so nested structured defaults use the same recursive freezing, import, and source path as canonical schemas. Added compile/import regressions across CLI, FastAPI, and MCP.

src/intpot/runtime.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. `App.tools` now always copies tool/parameter/list metadata and deep-copies defaults on a best-effort basis; non-copyable or self-copying opaque defaults remain available by identity. The compatibility contract and changelog now state that limitation explicitly, with lock and self-copying regressions.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. Canonical construction now rejects datetime/time defaults carrying `fold` or timezone semantics rather than silently losing them. Regression coverage includes folded and timezone-aware `datetime` and `time` values.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. Every non-JSON-native default now uses an unambiguous `$intpot` envelope that tags source container type; ordinary mappings are themselves wrapped, so list/tuple and bytes/lookalike-mapping values cannot collide. Strict JSON and non-collision regressions cover these cases.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. `_FrozenDict.__getitem__` now follows identity-or-equality lookup semantics, so an iterated NaN key is retrievable and `dict(mapping)` works. Added a direct Mapping regression.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. Decimal NaN defaults are rejected with a controlled `TypeError` because they cannot satisfy stable hashing/equality. Regression coverage includes signaling NaN construction and stable supported Decimal behavior.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. FastAPI parameter factories are rendered through dynamically allocated private aliases, so preserved `pathlib.Path` imports cannot overwrite the framework `Path` helper. Added a generated-module import regression combining a preserved Path import and path-sourced parameter.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. Generated-default dependencies now use dynamically allocated qualified module aliases chosen against every generated tool name and preserved-import binding, with deterministic suffixing on collision. Multi-tool compile/import regressions cover all helper names, preferred private aliases, and a preserved alias binding across all three generators.

src/intpot/core/models.py
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. `ParameterSchema` equality/hash now use a recursive source-type-sensitive default identity, distinguishing `bool` from `int` and container categories. The identity also preserves multiplicity for distinct NaN set members. Added bidirectional scalar and duplicate-NaN set regressions.

README.md
tugrulguner: Fixed in `9d03eedb9de5a6423bd85e20000d00080f0a8492`. README and generated Python skill now state that enum defaults are rejected, describe the `$intpot` envelope, and document opaque/temporal compatibility limits consistently with runtime behavior and the changelog.

tugrulguner (commented):
Exact-head adversarial re-review completed for `82925cb63f9d454a57f5f787649adfd0169eba4a`. The follow-up diff fixes wildcard import shadowing of generated framework/module/built-in aliases and preserves representation-sensitive Decimal, float, and complex schema identity. Focused adversarial coverage passed (16 tests) with no remaining findings.

tugrulguner:
Final follow-up fixes are live at `82925cb63f9d454a57f5f787649adfd0169eba4a`:

- wildcard source imports can no longer shadow generated framework, default-module, or built-in helpers
- schema equality preserves Decimal quantum/sign and float/complex signed zero
- regressions execute generated CLI, FastAPI, and MCP modules under hostile wildcard exports

Verification on this exact head: `make check` passed (**536 tests**; Ruff, formatting, and Pyright clean); generated CLI/MCP/FastAPI examples passed; fresh wheel and sdist isolated installs passed the wildcard/equality smoke; exact-head GitHub CI is green across Python 3.11–3.14, compatibility, lint, typecheck, build, and changelog; final adversarial re-review found no issues. Unresolved review threads: 0.

tugrulguner (commented):
## Exact-head fix re-review

No remaining findings. The changes in `9d03eedb9de5a6423bd85e20000d00080f0a8492` and `82925cb63f9d454a57f5f787649adfd0169eba4a` address all ten findings with targeted regression coverage.

### Verification
- `make check`: 536 passed; Ruff, format, and Pyright clean
- generated CLI, MCP, and FastAPI example behavior passed
- independent adversarial closure script reproduced and verified all ten prior cases
- all 22 review threads are resolved
- exact-head GitHub checks are green

Reviewed head: `82925cb63f9d454a57f5f787649adfd0169eba4a`

src/intpot/core/models.py
```diff
@@ -49,6 +61,576 @@ def __repr__(self) -> str:
...
+            _FrozenDict,
+            _FrozenFrozenset,
+            _FrozenList,
+            _FrozenSlice,
+            _FrozenTuple,
+            _FrozenSet,
+        ),
+    ):
+        return value
+    if isinstance(value, bytearray):
```
tugrulguner, line 203: **High — supported container subclasses are accepted but silently lose behavior.** These `isinstance` branches convert `defaultdict`/`Counter` to a plain dict and named tuples to a plain tuple. For example, `ParameterSchema('x', default=defaultdict(list)).to_info().default` is a `dict` with no default factory, and a named-tuple default loses its fields. The schema contract rejects values it cannot preserve, so container subclasses should either be handled explicitly or rejected; only exact built-in container types should enter these normalization branches. Add regressions for `defaultdict` and named tuple defaults.

src/intpot/core/models.py
```diff
@@ -49,6 +61,576 @@ def __repr__(self) -> str:
...
+    if isinstance(value, _FrozenFrozenset):
+        return ("frozenset", _multiset_identity(value))
+    if isinstance(value, _FrozenSlice):
+        return (
+            "slice",
+            _default_identity(value.start),
+            _default_identity(value.stop),
+            _default_identity(value.step),
+        )
+    if isinstance(value, float):
```
tugrulguner, line 621: **Medium — signed NaN semantics are lost across equality, JSON, and generated source.** `ParameterSchema('x', default=float('nan'))` and the same schema with `default=-float('nan')` compare equal and hash identically because both `hex()` values are `'nan'`; `to_dict()` emits the same tag, and `_source_default()` renders both as `float('nan')`. The negative input is therefore regenerated as a positive NaN even though `math.copysign()` distinguishes the stored defaults. Preserve the NaN sign/payload in the frozen identity and rendering, or reject NaNs if standalone generation cannot preserve them.

src/intpot/core/models.py
```diff
@@ -49,6 +61,576 @@ def __repr__(self) -> str:
...
+        return ("float", value.hex())
+    if isinstance(value, complex):
+        return (
+            "complex",
+            _default_identity(value.real),
+            _default_identity(value.imag),
+        )
+    if isinstance(value, Decimal):
+        return ("decimal", value.as_tuple())
+    return (type(value), value)
```
tugrulguner, line 631: **Medium — range schema identity uses sequence equality rather than constructor identity.** The fallback stores the `range` object directly in the equality key, so `range(0, 3, 2)` and `range(0, 4, 2)` produce equal schemas and hashes even though `to_dict()` and generated source preserve different `stop` values. Add a range-specific identity `(start, stop, step)` so equality matches the representation-sensitive contract used for other supported defaults.

src/intpot/core/models.py
```diff
@@ -158,3 +740,197 @@ def __post_init__(self) -> None:
...
+    http_method: str = "POST"
+    function_body: str | None = None
+    is_async: bool = False
+    route_path: str | None = None
+    dependencies: tuple[str, ...] = ()
+    source_imports: tuple[str, ...] = ()
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "name", sanitize_identifier(self.name))
+        object.__setattr__(self, "parameters", tuple(self.parameters))
```
tugrulguner, line 842: **High — direct canonical construction can generate duplicate function parameters.** Each `ParameterSchema` sanitizes independently, but `ToolSchema.__post_init__` only tuples the sequence and never applies `deduplicate_identifiers`. `ToolSchema(parameters=(ParameterSchema('a-b'), ParameterSchema('a_b')))` therefore stores two `a_b` parameters, and every generator emits `SyntaxError: duplicate argument 'a_b'`. `ToolInfo` already handles this collision; the public canonical constructor needs the same invariant or a deliberate validation error.

src/intpot/core/generators/_render.py
```diff
@@ -90,19 +92,67 @@ def _escape_docstring(text: str) -> str:
 }
 
 
-def _collect_extra_imports(tools: list[ToolInfo]) -> list[str]:
+def _private_aliases(tools: Sequence[RenderableTool]) -> dict[str, str]:
+    """Choose deterministic helper aliases that cannot collide with source globals."""
+    occupied = {tool.name for tool in tools}
```
tugrulguner, line 97: **High — tool names can still shadow generated runtime globals and later annotations.** Helper aliases avoid tool names, but the fixed `app`/`mcp` globals and annotation bindings do not. A first tool named `app` makes the next CLI/API decorator fail (`'function' object has no attribute 'command'/'post'`); a tool named `mcp` similarly breaks MCP import. A tool named `str` or `Any` also makes a later annotation resolve to that function rather than the intended type. Allocate collision-proof private names for all generated globals and separate public tool names from Python function identifiers (or reject reserved collisions), then execute multi-tool regressions across all targets.

src/intpot/converter.py
```diff
@@ -47,6 +53,50 @@ def inspect_app(source_type: SourceType, app_instance: Any) -> list[ToolInfo]:
...
+        name=_application_name(source_type, app_instance, source_path),
+        source_type=source_type,
+        tools=_inspect_tools(source_type, app_instance),
+        source_path=source_path,
+    )
+
+
+def inspect_app(source_type: SourceType, app_instance: Any) -> list[ToolInfo]:
+    """Inspect an app and return detached compatibility tool definitions."""
+    return compile_app(source_type, app_instance).to_tools()
```
tugrulguner, line 97: **Medium — the compatibility inspection APIs now reject opaque defaults they previously returned.** `inspect_app()` routes through `compile_app()`, and `IntpotApp.tools` routes through cached `.schema`, so an API endpoint with an opaque/un-copyable default raises `TypeError` before either public compatibility path can return `ToolInfo`. The underlying inspector still returns the value, and `App.tools` was explicitly fixed to preserve this compatibility edge. Keep `inspect_app()`/`IntpotApp.tools` on detached inspector models and reserve canonical validation for `.schema`, projection, and generation, or narrow the documented compatibility promise consistently.

tugrulguner (commented):
## Post-merge full-diff audit

A fresh review of the complete PR, including the reviewer-driven fix commits, found six remaining contract gaps: three high and three medium. These are new findings, not repetitions of the 22 resolved threads.

### Verification
- `make check`: 536 passed; Ruff, formatting, and Pyright clean
- generated CLI, MCP, and FastAPI examples executed successfully
- exact-head GitHub checks are green
- each finding below was reproduced against exact head `82925cb63f9d454a57f5f787649adfd0169eba4a`

Because the PR is already merged, this review is informational; the high-severity generated-code and value-preservation gaps should be fixed in follow-up work.

PR #131, opened 2026-09-08
fix: preserve CLI return control flow

tugrulguner:
## Summary

- preserve API and MCP `return` statements when generating CLI implementations
- let the existing Typer command wrapper print returned values without changing control flow
- retain source return annotations on generated implementation functions
- postpone generated CLI annotation evaluation so source-framework return types remain import-safe on Python 3.11–3.13
- add real generated-command coverage for early, loop, bare, async, and framework-specific returns
- exercise the behavior in the checked-in advanced API conversion example

## Verification

- `make check` on Python 3.11 (558 passed; Ruff, format, and Pyright passed)
- `bash scripts/demo.sh`
- `uv run --python 3.11 python scripts/verify_generated_examples.py`
- `uv run --python 3.11 python -m pytest tests/test_examples.py tests/test_roundtrip.py tests/test_transforms.py -q` (55 passed)
- `uv build`
- `uvx twine check dist/*`
- `make changelog-draft`


src/intpot/core/transforms.py
```diff
@@ -48,7 +49,9 @@ def _is_dict_type(return_type: str) -> bool:
 def _target_return_type(tool: ToolInfo, source: SourceType, target: SourceType) -> str:
     """Determine the correct return type for the target framework."""
     if target == SourceType.CLI:
-        return "None"
+        # The generated command wrapper returns None, but its implementation
+        # function preserves and prints the source function's returned value.
+        return tool.return_type
```
tugrulguner, line 54: Medium: retaining the source return annotation introduces an import-time regression on Python 3.11–3.13 because _collect_extra_imports still drops framework imports. Reproduced on Python 3.11 with a valid FastAPI endpoint: `from fastapi.responses import HTMLResponse`, `@app.get('/hello', response_class=HTMLResponse)`, `def hello() -> HTMLResponse: return '<h1>Hello</h1>'`. The source serves HTTP 200, but `intpot.load(source).to_cli()` now emits `_hello_impl() -> HTMLResponse` without importing HTMLResponse; executing the generated module raises `NameError: name 'HTMLResponse' is not defined`. With the base transform implementation, the same environment generates a CLI that exits 0 and prints the HTML. Python 3.14 lazy annotation evaluation hides the import-time failure. Preserve returns, but ensure retained implementation annotations have their required bindings (including framework return types), or use a safe annotation policy for the private implementation. Add a real generated-module import/invocation regression on an eagerly evaluated-annotation Python version.

tugrulguner (commented):
Exact-head review: one medium-severity regression requiring a fix before merge. Preserving return control flow is correct, but retaining implementation return annotations exposes missing framework imports in generated CLI modules. See inline reproduction. Verification: 45 focused transform/example tests passed; make check passed with 557 tests on Python 3.14; live CI is green (latest-dependencies skipped). Independently reproduced the import failure on Python 3.11 and confirmed the same source succeeds with the base transform implementation in the same environment. No code changes or merge performed. This author COMMENT review is not independent approval.

src/intpot/core/transforms.py
tugrulguner: Fixed in dec6ca4. Generated CLI modules now enable postponed annotation evaluation, so private implementation return annotations do not require source-framework bindings at import time. Added a real FastAPI `HTMLResponse` → CLI import/invocation regression using `compile(..., dont_inherit=True)` and reproduced RED then GREEN on Python 3.11. Verification: `make check` passed on Python 3.11 with 558 tests; focused generated-example/roundtrip/transform suite passed 55 tests.

tugrulguner (commented):
Exact-head re-review of dec6ca455b1c7c85fa39508e9e90a806d9ee378d: the reported import-time regression is fixed; no additional actionable findings in the reviewed changes. Generated CLI modules explicitly postpone annotations, and the regression compiles without inheriting the test module future flags. Independently regenerated the original HTMLResponse reproduction and executed it as a standalone file on Python 3.11: exit 0, output <h1>Hello</h1>. make check passed on Python 3.11: 558 tests, Ruff, formatting, and Pyright passed. Generated-example runtime verifier passed. Exact-head CI passes Python 3.11–3.14 and compatibility lanes; latest-dependencies is skipped. The original thread is resolved. No code changes or merge performed. This author COMMENT review is not independent formal approval.

PR #132, opened 2026-09-09
fix: preserve structurally required imports

tugrulguner:
## Summary

- replace substring-based source-import filtering with per-binding AST and symbol-table analysis
- retain globals required by generated bodies, annotations, defaults, aliases, forward references, augmented assignments, and deletions
- distinguish forward references from `Literal` values and `Annotated` metadata
- remove imports shadowed by parameters or locals and bindings whose framework-specific uses were translated away
- give generated framework imports collision-safe private aliases and reject unavoidable collisions with public generated bindings
- exercise generated CLI, API, and MCP artifacts and update the roadmap entry completed by #131

## Verification

- `make check` (575 tests; Ruff, formatting, and Pyright)
- focused generator, inspector, roundtrip, and generated-example suites on Python 3.11, 3.12, 3.13, and 3.14 (75 tests per version)
- `uv run python scripts/verify_generated_examples.py`
- `bash scripts/demo.sh` (idempotent regeneration)
- `uv build`
- `uvx twine check dist/*`
- `make changelog-draft`


src/intpot/core/generators/_render.py
```diff
@@ -118,6 +110,15 @@ def unique(base: str) -> str:
             "uuid",
         )
     }
+    for framework, names in {
+        "cli": ("asyncio", "typer"),
+        "mcp": ("FastMCP",),
+        "api": ("uvicorn",),
+    }.items():
+        for name in names:
+            aliases[f"{framework}:{name}"] = unique(
```
tugrulguner, line 120: Medium: allocate CLI helper aliases against parameter names too. A FastAPI endpoint `def echo(_intpot_cli_typer: str) -> str: return _intpot_cli_typer` produces a wrapper whose parameter shadows the new module alias. Invoking with hello raises AttributeError: 'str' object has no attribute 'echo'. The base-generated CLI exits 0 and prints hello. `_intpot_cli_asyncio` similarly collides with the async wrapper helper. Reserve all wrapper-scope bindings when choosing aliases; add sync/async generated-command regressions.

tugrulguner (commented):
Exact-head review: three medium-severity regressions require fixes before merge. Each was reproduced through load(source).to_cli() on Python 3.11: source FastAPI endpoints work, base-generated CLIs succeed, head-generated CLIs fail. Findings cover helper-parameter shadowing, augmented-assignment reads, and class-scope fallback reads. make check passed with 575 tests, Ruff, formatting, and Pyright; 75 focused tests passed on Python 3.14; saving demo passed without snapshot drift. Exact-head CI is green (latest-dependencies skipped). No code changes or merge performed. This author COMMENT review is not independent approval.

src/intpot/core/generators/_render.py
```diff
@@ -155,6 +194,148 @@ def _collect_extra_imports(
...
+        if isinstance(node, ast.Name) and isinstance(node.ctx, ast.Del)
+    }
+    referenced: set[str] = set()
+
+    def collect(table: symtable.SymbolTable) -> None:
+        referenced.update(
+            symbol.get_name()
+            for symbol in table.get_symbols()
+            if symbol.is_global()
+            and (symbol.is_referenced() or symbol.get_name() in deleted_names)
```
tugrulguner: Medium: an augmented assignment reads its target even when symtable does not mark it referenced. With `from math import pi`, the body `global pi; pi += 1; return 'done'` needs the initial import. On Python 3.11 pi is global/assigned but not referenced, so this filter removes its import and the generated CLI raises NameError. Source and base-generated CLI succeed. The added regression returns pi after updating it, masking the missing AugAssign read. Include scope-correct augmented-assignment target reads and test without a subsequent explicit read.

src/intpot/core/generators/_render.py
```diff
@@ -155,6 +194,148 @@ def _collect_extra_imports(
...
+        for node in ast.walk(syntax_tree)
+        if isinstance(node, ast.Name) and isinstance(node.ctx, ast.Del)
+    }
+    referenced: set[str] = set()
+
+    def collect(table: symtable.SymbolTable) -> None:
+        referenced.update(
+            symbol.get_name()
+            for symbol in table.get_symbols()
+            if symbol.is_global()
```
tugrulguner: Medium: class-local symbols can still read module imports. A source importing `from math import pi` with body `class C:
    pi = pi
return C.pi` works, but the generated CLI now raises NameError because pi is class-local in symtable and this filter drops its import. Class-body name lookup can fall back to globals before the class attribute is assigned; it is not function-local lookup. Independently reproduced via a FastAPI endpoint and load(source).to_cli(): base exits 0 printing 3.141592653589793; head exits 1. Preserve class-scope fallback reads and add a real generated-consumer regression.

src/intpot/core/generators/_render.py
```diff
@@ -155,6 +194,148 @@ def _collect_extra_imports(
     return sorted(source_imports) + sorted(default_imports)
 
 
+def _referenced_names(tools: Sequence[RenderableTool]) -> set[str]:
+    """Names that survive into generated bodies and annotations."""
+    referenced: set[str] = set()
+    for tool in tools:
+        if tool.function_body:
+            referenced.update(_body_global_names(tool))
```
tugrulguner, line 203: Medium — nested quoted annotations lose their required bindings. The inspector now extracts dependencies from nested annotations, but _body_global_names only uses symbol tables, which do not see names inside string annotations. Independently verified with a ToolInfo body containing `def inner(value: "Decimal"): return value` followed by `return typing.get_type_hints(inner)["value"].__name__`, and source_imports containing `from decimal import Decimal` and `import typing`. The base-generated CLI exits 0 and prints Decimal; the head drops the Decimal import and exits 1 with NameError. Parse nested annotation ASTs as well as outer signatures, accounting for their scopes, and add a generated-command regression. This is an additional finding from the completed independent audit, separate from the three findings in the main review.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in c0035cd. Generated CLI helper aliases now reserve every rendered parameter name, with executable sync and async command regressions covering `_intpot_cli_typer` and `_intpot_cli_asyncio` collisions.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in c0035cd. Scope-aware analysis now treats `AugAssign` targets as reads only when they resolve to an existing global or class fallback binding. The executable regression no longer masks the case with a later explicit read.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in c0035cd. Class bodies now use statement-order-aware fallback analysis, including assignments, loops, context managers, definition expressions, pattern captures, conditional control flow, annotation-only names, and exception-target cleanup. Runtime regressions exercise generated CLI consumers and guard against false retention after prior class-local binding.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in c0035cd. Nested function and class annotations are traversed structurally, including quoted forward references while excluding `Literal` values and `Annotated` metadata. The regression loads a real FastAPI source, generates a CLI, invokes it, and resolves the nested annotation with `get_type_hints`.

tugrulguner:
Follow-up exact-diff review found one additional class-scope control-flow case: bindings created only in a loop `else` suite were treated as definite even though `break` skips that suite. Fixed in 7606d5a with an executable FastAPI→CLI regression. Local verification remains green: 584 tests, focused generated-artifact suites on Python 3.11–3.14, Ruff, formatting, Pyright, and generated-example execution.

src/intpot/core/generators/_render.py
```diff
@@ -155,6 +195,582 @@ def _collect_extra_imports(
...
+def _special_scope_reads(tree: ast.AST) -> set[str]:
+    collector = _SpecialScopeReadCollector()
+    collector.visit(tree)
+    return collector.referenced
+
+
+def _nested_annotation_names(tree: ast.AST) -> set[str]:
+    referenced: set[str] = set()
+    for node in ast.walk(tree):
+        if isinstance(node, ast.arg) and node.annotation is not None:
```
tugrulguner, line 710: Medium — nested annotation dependencies must be resolved in their enclosing scope, not all added as globals. Reproduced a FastAPI source importing `from decimal import Decimal as endpoint`, then defining `def endpoint() -> int:` whose body is `endpoint = int; def inner(value: endpoint): return value; return inner(7)` (nested def on its own indented lines). The source returns 7. Before this fix, the unused module import was removed and the generated CLI printed 7. This new walk adds the locally bound endpoint as a global dependency, retains the unrelated Decimal import, and generation raises ValueError for a collision with the generated endpoint name. Track local/closure/class bindings when collecting annotation dependencies; add a negative-retention regression alongside the positive quoted-annotation test.

src/intpot/core/generators/_render.py
```diff
@@ -155,6 +195,582 @@ def _collect_extra_imports(
...
+        self._class_bound[-1] = previous
+
+    def visit_NamedExpr(self, node: ast.NamedExpr) -> None:
+        if self._in_class_scope() and isinstance(node.target, ast.Name):
+            name = node.target.id
+            if name not in self._class_bound[-1] and name in _loaded_ast_names(
+                node.value
+            ):
+                self.referenced.add(name)
+            self.visit(node.value)
```
tugrulguner, line 520: Medium — a visited assignment expression is not necessarily executed. Class-body BoolOp/IfExp nodes currently use generic traversal, so `unused = False and (pi := 0)` marks pi definitely bound even though the assignment never runs. With `from math import pi` and a function containing `class Values: unused = False and (pi := 0); value = pi` (separate class-body lines), then `return Values.value`, the source endpoint returns 3.141592653589793. The generated CLI drops the pi import and raises NameError. Merge possible binding states for short-circuit/conditional expressions, or conservatively avoid treating conditionally visited writes as definite bindings. Add an executable class-fallback regression for this case.

tugrulguner (commented):
Fix re-review: the four original reproductions now pass, but two medium-severity scope-analysis problems remain, described inline. Independently verified make check on Python 3.11: 584 passed, Ruff/format/Pyright clean. Generated-example runtime verifier passes. Exact-head CI passes across Python 3.11–3.14 and compatibility lanes, with latest-dependencies skipped. Both new source FastAPI endpoints execute correctly; one generated CLI is falsely rejected and the other raises NameError. Not ready to merge. No code changes or merge performed. Author COMMENT review, not independent approval.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in cc2266c. Nested annotation supplementation now collects only names introduced by quoted annotation strings; ordinary annotation expressions remain governed by the symbol table, so enclosing-function locals are not promoted to global import dependencies. Added an executable FastAPI-to-CLI negative-retention regression for the reported local `endpoint` case, alongside the existing positive quoted `Decimal` regression.

src/intpot/core/generators/_render.py
tugrulguner: Fixed in cc2266c. Class-scope binding analysis now merges execution paths for short-circuit boolean and conditional expressions, and handles later operands of chained comparisons as conditional. Added executable generated-CLI regressions for the reported `False and (pi := 0)` case plus conditional-expression and chained-comparison variants.

tugrulguner:
A final adversarial review found a mixed nested-annotation edge case: set subtraction conflated a direct local annotation occurrence with a same-named quoted forward reference. Fixed in 0c5aaaa by structurally collecting only quoted annotation positions while preserving `Literal` and `Annotated` exclusions. Added an executable FastAPI→CLI regression where `tuple[T, "T"]` needs the module import only for the quoted occurrence. Verification: 586 full-suite tests; 86 focused generated-artifact tests on each Python 3.11–3.14; Ruff, formatting, Pyright, and generated-example execution all pass.

tugrulguner (commented):
Fix verification at 0c5aaaab8a9251013c150a1cd974695eb1971365: both findings from the last review are fixed. Re-ran the exact standalone reproductions: locally bound nested annotation generates and prints 7; short-circuited class assignment preserves math.pi and prints 3.141592653589793. The four earlier reproductions also still pass, including typing.get_type_hints on a nested quoted Decimal annotation. Reviewed the new BoolOp, Compare, IfExp handling and mixed quoted/unquoted annotation regression coverage; no additional actionable findings in the fix diff. make check: 586 tests passed on Python 3.11; Ruff, formatting and Pyright clean. Generated-example runtime verifier passed. CI is green across Python 3.11–3.14 and compatibility lanes; latest-dependencies lane is skipped. Ready for the author to merge from this review. This is an author-account COMMENT review, not independent approval. No repository code changes or merge performed.

PR #133, opened 2026-09-12
fix: align live and ejected behavior

tugrulguner:
## Summary

- add executable CLI, FastAPI, and FastMCP live-versus-ejected parity coverage
- align unannotated parameter validation, canonical names/help, FastAPI metadata, unknown-method fallback, async return handling, and renamed recursive tools
- preserve parameters, local assignments, nested shadowing, replacement-name collisions, and explicit global declarations while adapting recursive bindings
- update generated examples, verification commands, changelog, and roadmap status

## Test plan

- [x] `make check` — 596 passed
- [x] generated-example import and execution
- [x] parity tests on Python 3.11, 3.12, 3.13, and 3.14 — 10 passed each
- [x] parity tests with FastMCP 2.0.0 — 10 passed
- [x] changelog draft and `git diff --check`


src/intpot/core/transforms.py
```diff
@@ -15,6 +15,32 @@
...
 
+class _NameLoadRewriter(ast.NodeTransformer):
+    """Rewrite references to a generated function binding."""
+
+    def __init__(self, original: str, replacement: str) -> None:
+        self.original = original
+        self.replacement = replacement
+
+    def visit_Name(self, node: ast.Name) -> ast.AST:
+        if isinstance(node.ctx, ast.Load) and node.id == self.original:
```
tugrulguner: P2 — rewrite only references that resolve to the original function binding. Matching every ast.Load with this spelling also rewrites parameters, local variables, comprehension variables and nested-scope bindings while leaving their declarations unchanged. Reproduced with `@app.tool()` / `def echo(echo: str) -> str: return echo.upper()`: the CLI template changes the return to `_echo_impl.upper()`, so live and base-generated CLIs print HELLO, while the PR-generated CLI raises AttributeError ('function' object has no attribute 'upper'). Also reproduced `@app.tool(name='renamed_echo')` / `def original(value: str) -> str: original = value; return original.upper()`: registration rewrites the local read to renamed_echo, causing the generated API to return HTTP 500 instead of the live/base HTTP 200 with HELLO, and its generated CLI fails too. Use scope/binding-aware rewriting, with the complete function signature available, and preserve shadowed names. Add executable parity tests for same-name parameters, local assignments, and nested scopes alongside the recursive happy path.

tugrulguner (commented):
Reviewed exact head 53210c39ac3d6889dc3eaa96e6f4c6161fb14b4c against base 2f3c8685cec1657362312ef285f15a87643e2997. One medium-severity regression found in the new self-reference rewrite; see inline reproduction. Not ready to merge. Verification: make check passed with 594 tests on Python 3.14.7; generated-example runtime verifier passed; CI checks pass across Python 3.11–3.14 and compatibility lanes, with latest-dependencies skipped. Both reproductions execute correctly against the base and live interfaces, but fail after PR ejection. No repository code changes or merge performed. Submitted as COMMENT from the author account, not independent approval.

src/intpot/core/transforms.py
tugrulguner: Fixed in `c0588d6c352771c7b6b0e21b847c56954cb784e7`. Replaced blanket `ast.Name` substitution with binding-aware aliasing: parameters, local assignments, nested shadowing, replacement-name/parameter collisions, and explicit `global` declarations retain valid semantics while recursive global/closure references resolve to the generated binding. Added executable live/ejected coverage across CLI, FastAPI, and FastMCP for the reported same-name parameter and renamed local cases, nested shadowing, nested recursion, a renamed tool whose generated name collides with its parameter, and explicit global recursion. Verification: `make check` passed with 596 tests; the parity suite passed 10 tests on each Python 3.11–3.14 and FastMCP 2.0.0.

tugrulguner (commented):
Fix verification at c0588d6c352771c7b6b0e21b847c56954cb784e7: the reported scope-rewrite regression is fixed. Independently reran the exact same-name parameter and renamed-tool local-variable reproductions: generated CLIs now print HELLO and the generated API returns HTTP 200 with HELLO, matching live and base behavior. Reviewed the binding-based replacement and added nested-shadowing, renamed recursion, and explicit-global recursion parity coverage. No new actionable findings in the fix diff. make check passed: 596 tests on Python 3.14.7; Ruff, formatting and Pyright clean. Generated-example runtime checks passed. Verification used the exact-head GitHub archive after git transport timed out; initialized a disposable local Git index to satisfy the docs tests that require tracked paths. CI is green across Python 3.11–3.14 and compatibility lanes; latest-dependencies skipped. Ready for the author to merge from this review. Author-account COMMENT, not independent approval. No product code edits or merge performed.

PR #137, opened 2026-09-17
feat: centralize target parameter placement

tugrulguner:
## Summary

- add an immutable target projection that makes parameter placement explicit for CLI, FastAPI, and FastMCP
- make live CLI/FastAPI builders and standalone generators share the placement resolver while preserving runtime-only opaque defaults
- transform canonical schemas directly and structurally share unchanged tools and parameters

## Focused invariant

For every target-projected schema, each parameter has one explicit target placement:

- CLI: required parameters are arguments; defaulted parameters are options
- FastAPI: declared body/query/header/path placement is preserved, with body as the default
- FastMCP: parameters record the single native MCP input placement; no adapter is needed because FastMCP has no competing parameter location

CLI and FastAPI live builders consume the shared choice where a framework location must be selected. All generators receive the target-projected schema before rendering. The canonical name, default, and description remain the authoritative values carried through that projection.

## Acceptance criteria

- [x] projection is immutable, idempotent, and structurally shares unchanged records
- [x] compatibility mappings retain explicit placement
- [x] live and ejected CLI behavior matches through a real Typer invocation
- [x] live and ejected FastAPI body/query/header/path behavior matches through real requests and OpenAPI
- [x] live and ejected FastMCP required/default behavior matches through real tool calls
- [x] live serving does not construct `ApplicationSchema`, so export serialization is not a prerequisite
- [x] generators normalize unprojected inputs to their target before rendering
- [x] live CLI and FastAPI preserve callable-to-canonical parameter mappings, including sanitized-name collisions and reordered compatibility metadata

## Verification

- `make check` — 617 passed; Ruff and Pyright passed
- `uv run pytest tests/test_runtime_builders.py -k 'sanitized_metadata or colliding_sanitized or reordered_contracts' -q` — 4 passed
- `uv run python scripts/verify_generated_examples.py` — CLI, MCP, FastAPI dependency route, and semantic schema verified
- `uv run python /tmp/intpot_pr137_mapping_probe.py` — async, positional-only, and bound sanitized mappings passed
- `uv build --out-dir /tmp/intpot-phase2-review-build-final` — wheel and sdist built at head `990fc7d`
- independent adversarial review of the remediation — passed after adding canonical-identity lookup and reordered-contract coverage

## Remaining Phase 2 work

This intentionally does not claim the whole shared-interface milestone. Follow-up slices still need to centralize target-specific name policy, required/default adaptation beyond placement, description adaptation, response policy, and default-value serialization/source rendering. ROADMAP.md records those as open.


src/intpot/runtime_builders.py
```diff
@@ -112,6 +113,31 @@ def build_typer_app(name: str, tools: list[RegisteredTool]) -> _typer.Typer:
...
+
+        @functools.wraps(restored)
+        def endpoint(*args: Any, **kwargs: Any) -> Any:
+            return restored(*args, **kwargs)
+
+        signature = inspect.signature(restored)
+        metadata = {parameter.name: parameter for parameter in info.parameters}
+        parameters = []
+        for parameter in signature.parameters.values():
+            contract = metadata[parameter.name]
```
tugrulguner: P2 — retain the original-to-canonical parameter-name mapping before requiring an exact metadata key. ParameterInfo sanitizes names (including collapsing consecutive underscores), but inspect.signature retains the callable's original names. For a valid registered tool `def show(user__id: str) -> str: return user__id.upper()`, metadata contains `user_id`, so this lookup raises KeyError('user__id') while constructing the CLI. The new `contracts[param_name]` lookup in _fastapi_endpoint has the same regression. Independently executed on the base: CLI prints HELLO and POST /show with JSON string hello returns HTTP 200 / HELLO. On this head, both live builders raise KeyError before serving. Preserve the callable-name mapping when resolving placement/defaults instead of indexing sanitized metadata by raw signature names, and add executable CLI/API tests for valid names altered by sanitization. Do not silently conflate distinct signature names that sanitize alike.

tugrulguner (commented):
Exact-head review: one P2 regression found in live CLI/API construction; not ready to merge. Reproduced against both this head and base 38d256bdaf2e0779a1eadc1edcd9c04cc2ee34df with verified imported package paths. make check: 613 tests passed on Python 3.14; Ruff, formatting and Pyright clean. Generated CLI/MCP/API/semantic-schema example runtime checks passed. CI is green, with latest-dependencies skipped. A second focused review found no additional concrete transform/projection defects; its extra differential probe was blocked and is not counted as verification. Git transport timed out, so local verification used exact-commit GitHub archives with a disposable Git index for documentation tests. No product code edits or merge. Author-account COMMENT review, not independent approval.

src/intpot/runtime_builders.py
tugrulguner: Fixed in `990fc7d`. Live builders now reconstruct the original-to-canonical parameter mapping with the same sanitization and deduplication rules used by `ToolInfo`, then resolve contracts by canonical identity rather than raw callable names or list position. This preserves names such as `user__id`, keeps colliding names distinct (`user__id` / `user_id`), and remains correct if mutable compatibility metadata is reordered. Added executable Typer and FastAPI regressions plus a reordered query/header contract regression. Verified with `make check` (617 passed), generated-example execution, package build, and an additional async/positional-only/bound-method probe.

tugrulguner (commented):
Fix verification at 990fc7da67b2a900c981cbed553853b6f4e5aead: the reported sanitized-parameter lookup regression is fixed. Independently reran the exact user__id reproduction against this head: live CLI prints HELLO and live API returns HTTP 200 with HELLO. Reviewed canonical-name reconstruction, deduplication, contract validation, and new tests covering unannotated parameters, colliding names and reordered metadata. No new actionable findings in the fix diff. make check passed: 617 tests on Python 3.14, Ruff/format/Pyright clean. Generated CLI/MCP/API/semantic-schema runtime checks passed. CI green across Python 3.11–3.14 and compatibility lanes; latest-dependencies skipped. Verification used the exact-commit GitHub archive and a disposable Git index for documentation tests. Ready for the author to merge from this review. Author-account COMMENT, not independent approval. No product code edits or merge performed.

